PR #6, opened 2022-09-28
Eflumerf/reorganization

eflumerf:


wesketchum (commented):
I think this looks ok -- not doing additional harm at least. I'll have a followup PR I think to clean up a few of the configs.

PR #12, opened 2022-09-29
configs from testing at EHN1 

wesketchum:
capturing config updates from testing at ehn1, including...
* dqm 'cern' implementation
* output_path --> output_paths
* add in felixcontroller config generations

I have not changed the hardware maps to only use one link yet, even though this is what we got running here today ... will test again with 10 links tomorrow and hopefully resolve that issue.


eflumerf (approved):
Added a few more instances of output_path -> output_paths correction

PR #13, opened 2022-09-30
Eflumerf/update documentation

eflumerf:


eflumerf:
In discussions with John, Pengfei and others, we want to move more towards each directory having its own documentation file.

PR #16, opened 2022-10-03
correct large scale configurations

mroda88:


mroda88:
This should be opened agaisnt he prep release branch, so I wrong PR

PR #17, opened 2022-10-03
Correct large scale conf

mroda88:


eflumerf (approved):
LGTM. large_scale_system.json should have correct outputs now.

PR #18, opened 2022-10-03
added json configs for HD coldbox

glehmannmiotto:
added a sub-directory with hardware map, pinning file and json configs for daq, card controller and wibs.


config/hd_coldbox_tests/flx_card_hd_coldbox.json
```diff
@@ -0,0 +1,10 @@
+{
+    "flxcardcontroller": {
+        "hardware_map_file": "HardwareMap_VST.txt",
```
wesketchum: This looks like the wrong hardware map file name?

config/hd_coldbox_tests/daq_hd_coldbox.json
```diff
@@ -0,0 +1,122 @@
...
+    "mlt_send_timed_out_tds": false,
+    "mlt_max_td_length_ms": 1000
+  },
+  "dataflow": {
+    "host_dfo": "np04-srv-012.cern.ch",
+    "apps": [
+      {
+        "app_name": "dataflow0",
+        "token_count": 10,
+        "output_paths": [ "/data0/test", "/data1/test/" ],
```
wesketchum: why are there two output paths defined here?

wesketchum (commented):
I think it looks good overall: noted two places where things seemed wrong. Would replace https://github.com/DUNE-DAQ/daq-systemtest/pull/15. 

If these were the exact configs run in coldbox tests, please indicate so and that it worked.

config/hd_coldbox_tests/flx_card_hd_coldbox.json
glehmannmiotto: this is a bug indeed! fixing it now.

config/hd_coldbox_tests/daq_hd_coldbox.json
glehmannmiotto: this was just to be able to push up rate and fragment size without being limited by disk writing.

wesketchum (approved):

PR #32, opened 2023-02-15
Feature/wibeth

glehmannmiotto:
This PR introduces the emulation testing of wibeth data and also uses the asset tools to retrieve the correct binary file for the different detector data formats.
It has been tested against nightly (+ readoutlibs [glm/fix_window_matching](https://github.com/DUNE-DAQ/readoutlibs/tree/glm/fix_window_matching)).

The PDS, TDE and DQM emulation runs are for now commented out.

### DIFF
--- a/integtest/readout_type_scan.py
+++ b/integtest/readout_type_scan.py
@@ -34,6 +34,11 @@ wib2_frag_params={"fragment_type_description": "WIB2",
                   "hdf5_source_subsystem": "Detector_Readout",
                   "expected_fragment_count": number_of_data_producers,
                   "min_size_bytes": 29000, "max_size_bytes": 30000}
+wibeth_frag_params={"fragment_type_description": "WIBEth",
+                  "fragment_type": "WIBEth",
+                  "hdf5_source_subsystem": "Detector_Readout",
+                  "expected_fragment_count": number_of_data_producers,
+                  "min_size_bytes": 7264, "max_size_bytes": 14456}
 pds_frag_params={"fragment_type_description": "PDS",
                  "fragment_type": "DAPHNE",
                  "hdf5_source_subsystem": "Detector_Readout",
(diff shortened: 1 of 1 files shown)


bieryAtFnal:
I created a software area based on the N23-02-20 nightly build, and I added this branch and the readoutlibs:glm/fix_window_matching branch.  The readout_type_scan.py test ran successfully, so that looked great.  I did have some questions out of curiosity, though...

- are there any risks associated with accepting any valid WIB2 or WIBEth data file that is stored in the assets system?
- what changed to cause the number of ProtoWIB or DUNE-WIB frames to alternate between N and N+1 (and therefore the Fragment size to alternate between two values)?  (previously, it seemed to always been the same value)
Thanks

glehmannmiotto:
As described in the PR of readoutlibs the time windowing is now inclusive, so any timestamp <timestamp of next wib frame is considered still valid. Since the timestamps of the requested windows are not necessarily exactly matching the wib timestamps (clock ticks x32 for wib2, x2048 for wibeth) there is a variability in how many frames fit a readout window , even if the window is constant size.
Giovanna

———————-
CERN EP/DT-DI


On 21 Feb 2023, at 03:37, bieryAtFnal ***@***.***> wrote:

﻿

I created a software area based on the N23-02-20 nightly build, and I added this branch and the readoutlibs:glm/fix_window_matching branch. The readout_type_scan.py test ran successfully, so that looked great. I did have some questions out of curiosity, though...

  *   are there any risks associated with accepting any valid WIB2 or WIBEth data file that is stored in the assets system?
  *   what changed to cause the number of ProtoWIB or DUNE-WIB frames to alternate between N and N+1 (and therefore the Fragment size to alternate between two values)? (previously, it seemed to always been the same value)
Thanks

bieryAtFnal (approved):
I've confirmed that the integtest runs without errors.
I now understand the reason for the slight variation in the WIB fragment size.
I'll defer my question about using any valid assets file until we have a discussion about unrelated asset file handling.

PR #40, opened 2023-04-10
Updated integtest/iceberg_real_hsi_test.py for dunedaq-v4.0.0

bieryAtFnal:
The changes to the integtest for this PR included ones to the timing configs (both in the timing DAQ session and the data-taking DAQ session), changes related to the connectivity server, and some other minor changes.  I also added a .gitignore file to ignore integtest temporary files.

I should say a few words about how the Connection Server startup is being handled.

There are checks in the iceberg_real_hsi_test.py script that determine whether an appropriate timing session is running and whether an instance of the Connection Server is running on the correct port.
If the timing session is not running, hints are printed to the console for how to start it.  Those hints include configuration snippets that will automatically start the ConnSrvr as part of the timing session.  This is necessary because the timing session needs to register itself with the ConnSrvr.

The port that the ConnSrvr uses is 15879.  This is determined in the following way:  In the timing session config, "connectivity_service_port" is set to 13579.  Port offsets are applied to that number using the following calculation:

- the timing session "partition number" value of 4 adds 2000 to the port (4 * 500)
- a specific offset of 300 is used for the Connection Server.

So, overall, 2300 is added to 13579 to get 15879.

Within the data-taking session configuration, we simply specify 15879 for the Connectivity Service (aka Connection Server) port, and it does not get any port offsets applied (thank goodness).

### DIFF
--- a/integtest/iceberg_real_hsi_test.py
+++ b/integtest/iceberg_real_hsi_test.py
@@ -39,15 +39,18 @@ triggercandidate_frag_params={"fragment_type_description": "Trigger Candidate",
 # Determine if the conditions are right for these tests
 we_are_running_on_an_iceberg_computer=False
 hostname=os.uname().nodename
-#if "iceberg01" in hostname or "protodune-daq02" in hostname:
 if "iceberg01" in hostname:
     we_are_running_on_an_iceberg_computer=True
-the_global_timing_partition_is_running=False
-username=os.environ.get('USER')
-for proc in psutil.process_iter(['pid', 'name', 'username']):
-    if proc.username() == username and "nanotimingrc" in proc.name():
-        the_global_timing_partition_is_running=True
-print(f"DEBUG: hostname is {hostname}, iceberg-computer flag is {we_are_running_on_an_iceberg_computer} and global-timing-running flag is {the_global_timing_partition_is_running}.")
+the_global_timing_session_is_running=False
+for proc in psutil.process_iter():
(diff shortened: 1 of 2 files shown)


integtest/iceberg_real_hsi_test.py
```diff
@@ -39,15 +39,18 @@
...
+the_global_timing_session_is_running=False
+for proc in psutil.process_iter():
+    if "nanotimingrc" in proc.name() and "iceberg-integtest-timing-session" in proc.cmdline():
+        the_global_timing_session_is_running=True
+try:
+  urllib.request.urlopen('http://localhost:15879').status
+  the_connection_server_is_running=True
+except:
+  the_connection_server_is_running=False
+print(f"DEBUG: hostname is {hostname}, iceberg-computer flag is {we_are_running_on_an_iceberg_computer}, global-timing-running flag is {the_global_timing_session_is_running}, connection-server-running flag is {the_connection_server_is_running}.")
```
eflumerf: Since you've made it independent of user, should this also print the username that the global timing session is running under?

eflumerf: the_global_timing_session_username=nobody
then, at L47, the_global_timing_session_username = proc.username()

eflumerf (approved):
Test runs correctly, however the hints provided incorrect information assuming that `daq-systemtest` was checked out in `sourcecode`

eflumerf (approved):
Verified that the updated code prints the correct path.

bieryAtFnal:
Thanks a lot, Eric, merging this now.

PR #51, opened 2023-07-06
Updated the config/long_window_readout files for the latest daqconf, …

bieryAtFnal:
…etc. changes


eflumerf (approved):
LGTM, config generates and runs (after changing output path, as noted in README)

PR #53, opened 2023-07-06
Tweaked the default_system_with_tpg.json configuration a bit to help …

bieryAtFnal:
…reduce the many error messages in the logfiles.

The changes include:

- switching to the wibeth playback-data file that has all zeroes and enabling the emulator mode that generates occasional pattern ADC data
- setting the slowdown factor to 1, since that is what we feel is appropriate for WIBEth data
- setting the TriggerActivity prescale to 25 so that a reasonable amount of TAs are produced from the TPs
- allowing the MLT to generate triggers with overlapping readout windows


mroda88:
Perfect, tested and it's running smoothly

mroda88 (approved):

PR #54, opened 2023-07-06
Move parameters as needed to ensure that most of the configs pass daq…

eflumerf:
…conf. clock_speed_hz was removed from configs as 62.5 MHz is now the default


eflumerf:
Also, "host_timing" was removed from the "timing" block, and it wasn't completely obvious where that information should go. I put it into the "hsi.host_timing_hsi" parameter, but a timing expert should probably take a look and revise before running the timing configs.

PR #56, opened 2023-07-11
Unify k8s and ssh configurations

plasorak:
This PR gets rid of the k8s-specific configuration and adds:
```
    "start_connectivity_service": false,
    "connectivity_service_host": "np04-srv-023",
    "connectivity_service_port": 30005,
    "k8s_rte": "release"
```
to the boot section of each configuration. This will use the connectivity service at NP04 and, if `--force-pm k8s` is provided to `daqconf_multiru_gen`, generate valid K8s configurations.

By default, these configurations will still use ssh.


glehmannmiotto:
I further reduced configs and made a change in large_scale_system.json for this.

glehmannmiotto (commented):
Those changes look good to me and already eliminate some duplication of json files to ease maintenance. All configs can now run in k8s or ssh mode.
Nevertheless, the test directory, which checks that all configs can be generated should be updated to use the --force-pm options and to not use the files that were removed.

mroda88 (approved):
Worked

PR #62, opened 2023-09-05
Integtest changes for new wib pattern generation

bieryAtFnal:
These changes are needed to keep TPG-related integtests working after the latest change in the WIB pattern generation code.

They also include the restoration of the TPG test in readout_type_scan.py.


eflumerf (approved):
LGTM

PR #65, opened 2023-09-27
Increased the configured number of DFO/TRB trigger-record slots

bieryAtFnal:
…in several integtests to avoid startup issues when occasional slowness in creating DataRequest Senders happens.

With the existing number of tokens, I occasionally see Trigger Inhibits in the 3ru_3df integtest.


eflumerf (approved):
LGTM integration tests pass

PR #67, opened 2023-10-10
JCF: DUNE-DAQ/daq-deliverables#122: make daq-systemtest an installabl…

jcfreeman2:
…e package

From https://github.com/DUNE-DAQ/daq-deliverables/issues/122


dingp:
This worked as expect. I saw `config` and `integtest` got installed into the `<workarea>/install/daq-systemtest/share` directory.

PR #75, opened 2023-11-15
Changes to the script for running multiple integtests

bieryAtFnal:
Including,

- moving it to a new directory so that it can be put into the PATH
- renaming it
- printing the summary information in the logfile
- searching for the integtest scripts in the appropriate SHARE env var


eflumerf (approved):
LGTM. Installed just `daqsystemtest` in a new area, and was able to run the script and have it pick up the integtests from the install directory.

PR #77, opened 2023-11-30
Added an option to the daqsystemtest_integtest_bundle.sh script for stop-on-failure

bieryAtFnal:
This will be useful in running the tests many times but stopping when an intermittent (rare) problem occurs.


bieryAtFnal:
closing this PR since I've included these changes in the PR that also has longer values for the tpset_min_latency_ticks.

PR #78, opened 2023-12-07
Extended tpset min latency in integtests and added an option to stop the bundle script on a failure

bieryAtFnal:
The changes in this PR specify a longer-than-default value for the tpset_min_latency_ticks configuration parameter in several integtests to try to allow those tests to run without errors.

There is also a change to the script that runs multiple integtests (the "bundle" script) to support an option to stop running individual tests when a failure occurs.

### DIFF
--- a/scripts/daqsystemtest_integtest_bundle.sh
+++ b/scripts/daqsystemtest_integtest_bundle.sh
@@ -10,12 +10,13 @@ Usage:
 "${script_name}" [option(s)]
 
 Options:
-    -h, --help
+    -h, --help : prints out usage information
     -s <DAQ session number (formerly known as partition number), default=1)>
     -f <zero-based index of the first test to be run, default=0>
     -l <zero-based index of the last test to be run, default=999>
     -n <number of times to run each individual test, default=1>
     -N <number of times to run the full set of selected tests, default=1>
+    --stop-on-failure : causes the script to stop when one of the integtests reports a failure
 """
     let counter=0
     echo "List of available tests:"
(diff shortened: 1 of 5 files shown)


bieryAtFnal:
Here are sample instructions for validating these changes on an AL9 DUNE DAQ computer:

```
setup_dbt latest
dbt-create -c -n NAFD23-12-14 14DecFDDevMinLatencyTweaks
cd 14DecFDDevMinLatencyTweaks/sourcecode
git clone https://github.com/DUNE-DAQ/daqsystemtest.git -b kbiery/extend_tpset_min_latency_in_integtests
cd ..

dbt-workarea-env
dbt-build -j 20
dbt-workarea-env

daqsystemtest_integtest_bundle.sh -s 2 --stop-on-failure
```

Success will be indicated by all of the tests passing, and if one of the tests happens to fail, then the "bundle" script should stop immediately after that.

scripts/daqsystemtest_integtest_bundle.sh
```diff
@@ -93,6 +99,14 @@ while [[ ${overall_loop_count} -lt ${overall_run_count} ]]; do
         fi
 
         let individual_loop_count=${individual_loop_count}+1
+
+        if [[ ${stop_on_failure} -gt 0 ]]; then
+            search_result=`tail -20 ${ITGRUNNER_LOG_FILE} | grep -i fail`
```
eflumerf, line 104: This seems slightly delicate. I wonder if the pytest return code could also be used.

eflumerf (approved):

PR #82, opened 2024-03-05
added a CRT datatype test to the readout_type_scan integtest

bieryAtFnal:
This change depends on changes in many other repos, and it should not be merged until after all of those changes are merged.  For a list of those other repos, please see [PR 162 in fdreadoutlibs](https://github.com/DUNE-DAQ/fdreadoutlibs/pull/162).

The goal of this change was to simply add the CRT datatype to the list of ones that have a basic sanity check as part of our integtests.

This PR is part of the work associated with [daq-deliverables Issue 125](https://github.com/DUNE-DAQ/daq-deliverables/issues/125).

### DIFF
--- a/integtest/readout_type_scan.py
+++ b/integtest/readout_type_scan.py
@@ -22,16 +22,6 @@ expected_number_of_data_files=1
 check_for_logfile_errors=True
 expected_event_count=run_duration
 expected_event_count_tolerance=2
-wib1_frag_hsi_trig_params={"fragment_type_description": "WIB", 
-                           "fragment_type": "ProtoWIB",
-                           "hdf5_source_subsystem": "Detector_Readout",
-                           "expected_fragment_count": number_of_data_producers,
-                           "min_size_bytes": 37192, "max_size_bytes": 37656}
-wib1_frag_multi_trig_params={"fragment_type_description": "WIB",
-                             "fragment_type": "ProtoWIB",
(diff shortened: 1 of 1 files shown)


wesketchum:
So, I've merged this with the patch/fddaq-v4.4.x branch, and then pulled this down and tried to run it with all the other CRT branches. Two things:

- About 50% of the time the PDS readout scan fails due to a segmentation fault. On a second run, it's fine.
- The CRT scan gave me a number of 'Readout timeouts'. I'm not sure if there's a problem with the time assignment or what there ... but am also wondering if I have things properly pulled down / configured. We are running with CRT right now and things are looking ok ...

bieryAtFnal:
I was able to get the CRT part of the `readout_type_scan` to run without warnings by increasing the `crt_time_tick_diff` value from 16 to 1024 in [FDFakeCardReader](https://github.com/DUNE-DAQ/fdreadoutmodules/blob/ec77682639a59b1cefc330ac52f999c00c653109/plugins/FDFakeCardReader.cpp#L109).

I have not committed that change, but I believe that it will help.

I have not see the problems with the PDS part of the `readout_type_scan` that were mentioned above...

wesketchum:
confirm with update as described in fdreadoutmodules this works. Approving and merging for the patch release (that will have that too)

wesketchum (approved):

PR #89, opened 2024-04-22
 Updated the readout_type_scan integtest to take into account the new DAPHNE frame size

bieryAtFnal:
plus a minor change in the `tpstream_writer` integtest, and an update to the package version number.

With these changes, I'm seeing all of the integtests in the `daqsystemtest` package completing successfully.


wesketchum (approved):
LGTM

PR #101, opened 2024-07-02
Fixes for integtests on the v4 development line

bieryAtFnal:
Two categories of changes:

- updates to the expected PDS fragment sizes in the readout_type_scan.  This may need to change again after we discuss what the correct time-tick value should be used in emulated-data tests.  (The changes in the DAPHNE emulated data were triggered when the number of frames per superchunk was changed from 12 to 1.)
- the addition of a small wait time at the start of a run before enabling triggers when TPG is enabled.  This is to allow TPs to appear in all of the relevant latency buffers now that we have 3 times as many TP DataLinkHandlers (now one per plane).

### DIFF
--- a/integtest/3ru_1df_multirun_test.py
+++ b/integtest/3ru_1df_multirun_test.py
@@ -136,9 +136,9 @@ confgen_arguments={"WIBEth_System": conf_dict,
 # The commands to run in nanorc, as a list
 if sufficient_resources_on_this_computer:
     nanorc_command_list="integtest-partition boot conf".split()
-    nanorc_command_list+="start_run --wait 1 101 wait ".split() + [str(run_duration)] + "stop_run          wait 2".split()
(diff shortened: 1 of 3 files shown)


bieryAtFnal:
This PR, along with [fdreadoutmodules 21](https://github.com/DUNE-DAQ/fdreadoutmodules/pull/21), is ready for review.

As described in the `fdreadoutmodules` PR, to test it, I would recommend creating a new software area using [these instructions](https://github.com/DUNE-DAQ/daqconf/wiki/Instructions-for-setting-up-an-fddaq%E2%80%90v4.4.4-software-area), then update the `daqsystemtest` repo to branch `kbiery/v4.4.x_fixes_for_integtests` and the `fdreadoutmodules` repo to branch `kbiery/daphne_time_tick_update`.  Rebuild and then run `daqsystemtest_integtest_bundle.sh -s 2 --stop-on-fail`.  All tests should pass.

wesketchum (approved):

PR #106, opened 2024-07-22
Updated the expected range of DAPHNE fragment sizes in readout_type_s…

bieryAtFnal:
…can to reflect the superchunk size change to 3


wesketchum (approved):
looks good.

PR #109, opened 2024-07-25
Increased the expected number of TimeSlices in the tpstream_writing_test

bieryAtFnal:
This is to take into account the recent addition of a little more wait time at the start of runs in this testing.  

When I added the extra wait time at the start of the run, I should have increased the expected number of TimeSlices.  The change in this PR takes care of that oversight.

In my original (and current) testing, I didn't see any problems with the tpstream_writing_test, but I think that I was just getting lucky.  The number of TimeSlices in the TPStream data files did increase, but it was just luckily staying within the allowed range.  I suspect that Michal saw it fluctuate outside the allowed range, and we should make the test as robust as possible in any case.

### DIFF
--- a/integtest/tpstream_writing_test.py
+++ b/integtest/tpstream_writing_test.py
@@ -197,7 +197,7 @@ def test_data_files(run_nanorc):
 
 def test_tpstream_files(run_nanorc):
     tpstream_files = run_nanorc.tpset_files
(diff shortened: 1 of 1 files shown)


MRiganSUSX (approved):
I rerun this test few times with the latest release and did not observe the error I've previously seen. 
LGTM.

wesketchum:
Going to merge this in so it can be part of the patch release.

PR #113, opened 2024-09-18
add file URI for opmon

mroda88:


eflumerf (approved):

eflumerf:
Do you want to change the opmon_uri of the local sessions?

eflumerf:
https://github.com/DUNE-DAQ/daqsystemtest/blob/develop/config/oks_configs/example-configs.data.xml#L165

Also, you can remove it from the integrationtest-objects.data.xml:
https://github.com/DUNE-DAQ/daqsystemtest/blob/develop/config/daqsystemtest/integrationtest-objects.data.xml#L96C1-L100C1

PR #115, opened 2024-09-20
Removed the session number from the daqsystemtest_integtest_bundle.sh script…

bieryAtFnal:
… now that we are using drunc for integtests, and we get errors when we try to pass --nanorc-options to integtests.


eflumerf (approved):
I will note, however, that `--nanorc-option log-level debug` can be useful for figuring out why the drunc-controllers died suddenly during/after boot.

PR #120, opened 2024-10-01
Set the request_timeout to zero for the sample config HSI request handler

bieryAtFnal:
Updated the HSI request handler sample configuration to have a request_timeout value of zero and not print out warnings when data requests time out.  This is to avoid unnecessary waiting for Fragments at the TRB.

To test this, one can use the following command to run a small system:
```
drunc-unified-shell ssh-standalone config/daqsystemtest/example-configs.data.xml local-1x1-config boot wait 5 conf wait 3 start 101 enable-triggers wait 10 disable-triggers drain-dataflow stop-trigger-sources stop scrap terminate
```

Without this change, you will see trigger inhibits being generated (and they will be mentioned in the log files).  With this change, there will be no inhibits.  (There is more detailed testing that one could do, such as looking at TRACE debug messages, and I can provide instructions if anyone is interested.)

### DIFF
--- a/config/daqsystemtest/hsi-segment.data.xml
+++ b/config/daqsystemtest/hsi-segment.data.xml
@@ -163,10 +163,10 @@
 
 <obj class="RequestHandler" id="hsi-requesthandler-generic">
  <attr name="handler_threads" type="u16" val="4"/>
- <attr name="request_timeout" type="u32" val="1000"/>
+ <attr name="request_timeout" type="u32" val="0"/>
  <attr name="pop_limit_pct" type="float" val="0.8"/>
  <attr name="pop_size_pct" type="float" val="0.1"/>
- <attr name="warn_on_timeout" type="bool" val="1"/>
+ <attr name="warn_on_timeout" type="bool" val="0"/>
  <attr name="warn_on_empty_buffer" type="bool" val="0"/>
 </obj>
 


eflumerf (approved):

eflumerf:
There may need to be an associated change in `daqconf` for generated HSI segments

bieryAtFnal:
Good point.

Is there an automated regression test that still uses the FakeHSI?  (so that I/we can take a look at its generated configuration)

Also, I don't see any instances of the string "request_timeout" in the head of develop of daqconf.  So, that means that we would be adding the setting of the request_timeout to the generate_hsiOKS script, rather than modifying it, correct?

bieryAtFnal:
Ah, I see that small_footprint_quick_test uses the fake HSI...
I'll look at it...

bieryAtFnal:
Is our discovery that the default values set in the hsi-requesthandler-generic object in integrationtest-objects.data.xml ([here](https://github.com/DUNE-DAQ/daqsystemtest/blob/955f32b185cfa3234f0af3a1d2ba7278c4f080f6/config/daqsystemtest/integrationtest-objects.data.xml#L130)) include a zero value for the request_timeout sufficient do address the request to include this change in HSI configuration genration?

eflumerf:
I would say so, yes

PR #121, opened 2024-10-01
Add the HSI and a deeper latency buffer to minimal system quick test

plasorak:
This seems to be the only way to not get the MLT to look for TC data sources (which is guaranteed to be >1 by the configuration),
Without this, the `IOManager` runs in a tight loop looking for the TC sources (overwhelming the connectivity server and the MLT).
With this one, I get inhibits, and increasing `data_rate_slowdown_factor` doesn't help. I'm missing something.

### DIFF
--- a/integtest/minimal_system_quick_test.py
+++ b/integtest/minimal_system_quick_test.py
@@ -88,13 +88,30 @@ conf_dict.dro_map_config.n_streams = number_of_data_producers
 conf_dict.op_env = "integtest"
 conf_dict.session = "minimal"
 conf_dict.tpg_enabled = False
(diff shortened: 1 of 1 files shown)


bieryAtFnal:
Would another possibility be to send the output of the RandomTriggerCandidateMaker to the data-reader-mlt/DataSubscriberModule?

I'll attach a picture of the current situation...
Screenshot 2024-10-01 at 2 18 59 PM

eflumerf:
We have implemented the alternate architecture so that RandomTCMaker can be used in minimal_system_quick_test.

PR #122, opened 2024-10-03
Changed the number of handler_threads from 0 to 1 in the hsi-requesth…

bieryAtFnal:
…andler-generic declaration in integrationtest-objects.data.xml.  This is to get the small_footprint_quick_test (and any other integtests that use the FakeHSI) to work.  And, it seems reasonable to have at least one data handler thread as a default.

### DIFF
--- a/config/daqsystemtest/integrationtest-objects.data.xml
+++ b/config/daqsystemtest/integrationtest-objects.data.xml
@@ -128,7 +128,7 @@
 </obj>
 
 <obj class="RequestHandler" id="hsi-requesthandler-generic">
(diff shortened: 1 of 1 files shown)


bieryAtFnal:
For reference, I found the problem using the `print_detailed_config_info` script.  I compared the hsi-01 configuration for the `small_footprint_quick_test` with the `local-1x1-config` (in `daqsystemtest`), and noticed that the number of `handler_threads` was zero in the `small_footprint_quick_test` case.

eflumerf (approved):
Verified that `small_footprint_quick_test` now passes

PR #125, opened 2024-10-08
candidate changes to get the long_window_readout test to work again

bieryAtFnal:
These changes restore the focus on the split-TR test, for now, and add the --trigger-rate start-run parameter to get the right trigger rate set up.


eflumerf (approved):
The test now passes

PR #129, opened 2024-10-14
Initial tweaks to ignored errors and warnings in integtests after new…

bieryAtFnal:
…ly appearing and modified messages.


eflumerf (approved):
In general, these exclusions are probably better than the ones I was using.

PR #133, opened 2024-10-16
No thread pinning

plasorak:
Remove the thread pinning on the example configurations.

### DIFF
--- a/config/daqsystemtest/fsm.data.xml
+++ b/config/daqsystemtest/fsm.data.xml
@@ -386,7 +386,6 @@
  </attr>
  <attr name="mandatory" type="string">
   <data val="user-provided-run-number"/>
-  <data val="file-run-registry"/>
  </attr>
 </obj>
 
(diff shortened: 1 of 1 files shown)


bieryAtFnal (approved):
From the "biery" account on np04-srv-018 and np04-srv-003, the changes in PR 133 help with the running of the minimal_system_quick_test.  And, they don't seem to hurt anything from the biery account on [daq.fnal.gov](http://daq.fnal.gov/).
From the np04daq account on np04-srv-003, they help with the running of the minimal_system_quick_test, but only once I unset the web proxy.
Also, from the np04daq account on np04-srv-003, they don't help with interactive running of drunc with the local-1x1-config.  In that case, all of the controller processes seem to exit immediately, and don't print anything in the logs.  Maybe that is a separate problem?

plasorak:
> Also, from the np04daq account on np04-srv-003, they don't help with interactive running of drunc with the local-1x1-config. In that case, all of the controller processes seem to exit immediately, and don't print anything in the logs. Maybe that is a separate problem?

Sounds like a different problem yes, I'll have a look

PR #134, opened 2024-10-16
Since drunc PR #266 has been merged, we can remove a number of log-file

eflumerf:
exclusion strings from the integration tests.


bieryAtFnal (approved):
Looks great to me.
I've confirmed that the integtests no longer need these exclusions (they do not complain about them now that we are using the modified drunc).

PR #142, opened 2024-10-24
Reduce the chatter of the connectivity service

plasorak:


eflumerf (approved):
Surprised this was not already implemented. Note that the CONNECTION_FLASK_DEBUG environment variable enables better log messages anyway.

eflumerf:
Compare with: https://github.com/DUNE-DAQ/integrationtest/blob/develop/python/integrationtest/integrationtest_drunc.py#L256

PR #144, opened 2024-10-28
Add two potential parameters for controlling the ConnectivityService

eflumerf:
startup and listening.


bieryAtFnal (approved):
Looks fine to me.

PR #151, opened 2024-11-05
Work around https://github.com/DUNE-DAQ/trigger/issues/357 by setting

eflumerf:
the HSISignalWindow time_before to be equal to that of the relevant TCReadoutMap.

### DIFF
--- a/config/daqsystemtest/moduleconfs.data.xml
+++ b/config/daqsystemtest/moduleconfs.data.xml
@@ -63,7 +63,7 @@
 
 <oks-data>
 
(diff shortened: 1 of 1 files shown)


bieryAtFnal (approved):
I've confirmed that with this change, the following command produces TriggerRecords that all have non-empty TriggerCandidate fragments.
```
drunc-unified-shell ssh-standalone config/daqsystemtest/example-configs.data.xml local-1x1-config boot --no-override-logs wait 5 conf wait 3 start 112 change-rate 1.0 enable-triggers wait 30 disable-triggers drain-dataflow stop-trigger-sources stop scrap terminate
```

PR #152, opened 2024-11-05
Workarounds for Trigger #357

eflumerf:
Note that after this PR is merged, small_footprint_quick_test will fail until https://github.com/DUNE-DAQ/trigger/issues/357 is resolved.


bieryAtFnal (approved):
LGTM

PR #155, opened 2024-11-07
Updating start <run_no> -> start --run-number <run_no>

PawelPlesniak:
Requires https://github.com/DUNE-DAQ/drunc/pull/292
Now in the tests, when running `start`, you need to specify `start --run-number`


plasorak:
Works with the drunc PR

PR #157, opened 2024-11-11
Added a tmp_path_retention_count value of 10 to the pytest.ini file s…

bieryAtFnal:
…o that we keep all of the tmpdirs from a full daqsystemtest_integtest_bundle.sh run (plus 1 pre-existing one).


eflumerf (approved):
Verified that I am now retaining more pytest directories. Note that on some systems, this may fill up /tmp when running lots of tests with large output files.

PR #162, opened 2024-11-19
Unify all `<app_id>_control`

plasorak:
This PR uses https://github.com/DUNE-DAQ/confmodel/pull/53 to make the application configuration simpler, by removing all the `<app_id>_control` services and unifying them in `daqapp_control` and `drunc_control`.

PR #164, opened 2024-11-20
Unify all `<app_id>_control` for `develop`

plasorak:
Same as https://github.com/DUNE-DAQ/daqsystemtest/pull/162, for `develop`.


eflumerf (approved):
Had to check out `appmodel` for testing. Works after merging develop

PR #167, opened 2024-11-26
Expose ta-postprocessing-disabling option in trigger schema

ArturSztuc:
Explained in PR https://github.com/DUNE-DAQ/appmodel/pull/159


ArturSztuc:
Closing this PR. Changed variable to `tx_generation_disabled`, it's false by default, and we probably don't want to keep it in our configuration explicitly set to default since it's only for the experts to change...

PR #168, opened 2024-12-09
Add TP Time Over Threshold Minima Configurations.

aeoranday:
Updating the test configuration to be consistent with changes in https://github.com/DUNE-DAQ/appmodel/pull/161.

I've placed the change in `moduleconfs.data.xml`, but I'm not confident that this is necessarily the correct file for it.


glehmannmiotto (approved):
LGTM

PR #172, opened 2024-12-31
Added back the ability to specify a non-standard output directory for…

bieryAtFnal:
… data files in a couple of integtests.

This ability was lost in the transition from v4 software to v5 software.

This PR depends on one in the integrationtest repo ([PR 95](https://github.com/DUNE-DAQ/integrationtest/pull/95)).

To test these changes, one needs to create a software area that includes both the changes in the `integrationtest` and `daqsystemtest` repos, and one needs to specify a non-standard value for the `output_path_parameter` local variable in the `long_window_readout_test` and/or the `output_dir` local variable in the `tpstream_writing_test`.


eflumerf (approved):
Changing output path now works.

PR #173, opened 2025-01-07
Add method to check Fragment error flags

eflumerf:
Uses https://github.com/DUNE-DAQ/integrationtest/pull/96

Note that TriggerCandidate Fragments currently all have the "kIncomplete" error flag set. This should be resolved or masked before merging. Also only minimal_system_quick_test and 3ru_3df_multirun_test currently call the method.


bieryAtFnal (approved):
LGTM.
As we've discussed in Core SW meetings and on Slack, we need to have a future discussion on how to handle status/error bits in fragments, especially in light of the fact that TriggerCandidate fragments currently always have the kIncomplete bit set.  In the meantime, we are moving ahead with the infrastructure changes and focusing the checking of fragment error flags in fragments in integtests to non-Trigger fragments.

PR #177, opened 2025-01-24
Many ELisAs

plasorak:
To be used with https://github.com/DUNE-DAQ/drunc/pull/344


PawelPlesniak (approved):
Tested with https://github.com/DUNE-DAQ/drunc/pull/344

PR #186, opened 2025-03-06
TP Replay Application (v5)

MRiganSUSX:
Details here: https://github.com/DUNE-DAQ/trigger/pull/382

### DIFF
--- a/config/daqsystemtest/ccm.data.xml
+++ b/config/daqsystemtest/ccm.data.xml
@@ -63,7 +63,7 @@
 
 <oks-data>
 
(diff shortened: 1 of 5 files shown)


ArturSztuc (approved):
LGTM.
I played with this configuration and it generally worked as expected, though wasn't able to run all the tests I wanted.

I'm now thinking (not for this PR!) it would be good to have an integration test that involves the replay application. That way it would be faster/easier to spot e.g. changes in the trigger (or other parts of the DAQ) that are breaking the replay application.

ArturSztuc (approved):
LGTM, though the change of the config order has threw me off initially :D

PR #189, opened 2025-03-12
Adding debug variables

MRiganSUSX:
I find these variables useful when developing / debugging. 
Is there a reason they should not be part of the pre-defined set?

### DIFF
--- a/config/daqsystemtest/ccm.data.xml
+++ b/config/daqsystemtest/ccm.data.xml
@@ -224,7 +224,7 @@
 
 <obj class="Variable" id="local-env-dunedaq-ers-debug-level">
  <attr name="name" type="string" val="DUNEDAQ_ERS_DEBUG_LEVEL"/>
- <attr name="value" type="string" val="0"/>
+ <attr name="value" type="string" val="-1"/>
 </obj>
 
 <obj class="VariableSet" id="ehn1-variables">
--- a/config/daqsystemtest/ccm.data.xml
+++ b/config/daqsystemtest/ccm.data.xml
@@ -224,7 +224,7 @@
 
 <obj class="Variable" id="local-env-dunedaq-ers-debug-level">
  <attr name="name" type="string" val="DUNEDAQ_ERS_DEBUG_LEVEL"/>
- <attr name="value" type="string" val="-1"/>
+ <attr name="value" type="string" val="0"/>
 </obj>
 
 <obj class="VariableSet" id="ehn1-variables">
@@ -245,6 +245,7 @@
   <ref class="Variable" id="local-env-ers-info"/>
   <ref class="Variable" id="local-env-ers-verb"/>
   <ref class="Variable" id="local-env-ers-warning"/>
+  <ref class="Variable" id="local-env-dunedaq-ers-debug-level"/>
  </rel>
 </obj>
 
--- a/config/daqsystemtest/ccm.data.xml
+++ b/config/daqsystemtest/ccm.data.xml
@@ -222,13 +222,8 @@
  <attr name="value" type="string" val="erstrace,throttle,lstdout"/>
 </obj>
 
-<obj class="Variable" id="local-env-tlog-debug">
+<obj class="Variable" id="local-env-dunedaq-ers-debug-level">
  <attr name="name" type="string" val="DUNEDAQ_ERS_DEBUG_LEVEL"/>
- <attr name="value" type="string" val="0"/>
-</obj>
-
-<obj class="Variable" id="trace-lvl">
- <attr name="name" type="string" val="TRACE_LVLS"/>
  <attr name="value" type="string" val="-1"/>
 </obj>
 
--- a/config/daqsystemtest/ccm.data.xml
+++ b/config/daqsystemtest/ccm.data.xml
@@ -63,7 +63,7 @@
 
 <oks-data>
 
-<info name="" type="" num-of-items="27" oks-format="data" oks-version="862f2957270" created-by="gjc" created-on="thinkpad" creation-time="20231207T110859" last-modified-by="eflumerf" last-modified-on="ironvirt9.mshome.net" last-modification-time="20241120T175651"/>
+<info name="" type="" num-of-items="29" oks-format="data" oks-version="862f2957270" created-by="gjc" created-on="thinkpad" creation-time="20231207T110859" last-modified-by="mrigan" last-modified-on="np04-srv-016.cern.ch" last-modification-time="20250312T105152"/>
 
 <include>
  <file path="schema/confmodel/dunedaq.schema.xml"/>
@@ -75,6 +75,7 @@
  <comment creation-time="20240208T094131" created-by="eflumerf" created-on="ironvirt9.mshome.net" author="eflumerf" text="reformat"/>
  <comment creation-time="20240909T081245" created-by="glehmann" created-on="np04-srv-024.cern.ch" author="glehmann" text=" "/>
  <comment creation-time="20240918T094920" created-by="maroda" created-on="np04-srv-015.cern.ch" author="maroda" text="add local file"/>
+ <comment creation-time="20250312T102000" created-by="mrigan" created-on="np04-srv-016.cern.ch" author="mrigan" text=" "/>
 </comments>
 
 
@@ -219,6 +220,16 @@
 <obj class="Variable" id="local-env-ers-warning">
  <attr name="name" type="string" val="DUNEDAQ_ERS_WARNING"/>
  <attr name="value" type="string" val="erstrace,throttle,lstdout"/>
+</obj>
+
+<obj class="Variable" id="local-env-tlog-debug">
+ <attr name="name" type="string" val="DUNEDAQ_ERS_DEBUG_LEVEL"/>
+ <attr name="value" type="string" val="0"/>
+</obj>
+
+<obj class="Variable" id="trace-lvl">
+ <attr name="name" type="string" val="TRACE_LVLS"/>
+ <attr name="value" type="string" val="-1"/>
 </obj>
 
 <obj class="VariableSet" id="ehn1-variables">
--- a/config/daqsystemtest/ccm.data.xml
+++ b/config/daqsystemtest/ccm.data.xml
@@ -222,7 +222,7 @@
  <attr name="value" type="string" val="erstrace,throttle,lstdout"/>
 </obj>
 
-<obj class="Variable" id="local-env-dunedaq-ers-debug-level">
+<obj class="Variable" id="local-env-ers-debug-level">
  <attr name="name" type="string" val="DUNEDAQ_ERS_DEBUG_LEVEL"/>
  <attr name="value" type="string" val="-1"/>
 </obj>
@@ -245,7 +245,7 @@
   <ref class="Variable" id="local-env-ers-info"/>
   <ref class="Variable" id="local-env-ers-verb"/>
   <ref class="Variable" id="local-env-ers-warning"/>
-  <ref class="Variable" id="local-env-dunedaq-ers-debug-level"/>
+  <ref class="Variable" id="local-env-ers-debug-level"/>
  </rel>
 </obj>
 
--- a/config/daqsystemtest/example-configs.data.xml
+++ b/config/daqsystemtest/example-configs.data.xml
@@ -144,6 +144,7 @@
   <ref class="Variable" id="local-env-ers-warning"/>
   <ref class="Variable" id="local-env-ers-error"/>
   <ref class="Variable" id="local-env-ers-fatal"/>
+  <ref class="Variable" id="local-env-dunedaq-ers-debug-level"/>
  </rel>
  <rel name="disabled">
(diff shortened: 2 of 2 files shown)


MRiganSUSX:
For context:
- the debug variable allows TLOG_DEBUG messages to be shown in log file, for example using `DUNEDAQ_ERS_DEBUG_LEVEL=10`
will result in messages of type:
```
2025-Mar-12 11:40:46,637 DEBUG_3 [void dunedaq::trigger::TCProcessor::make_td(const dunedaq::trigger::TCWrapper*) at /tmp/root/spack-stage/spack-stage-trigger-NB_DEV_250312_A9-673ugf5vzrdrgpaziba63uhcaro6z23h/spack-src/src/TCProcessor.cpp:227] Got TC of type 1, timestamp 108861002914740380, start/end 108861002914737380/108861002914760380, readout start/end 108861002914737380/108861002914741381

2025-Mar-12 11:40:47,304 DEBUG_10 [dunedaq::dfmessages::ComponentRequest dunedaq::trigger::TCProcessor::create_request_for_link(dunedaq::dfmessages::SourceID, triggeralgs::timestamp_t, triggeralgs::timestamp_t) at /tmp/root/spack-stage/spack-stage-trigger-NB_DEV_250312_A9-673ugf5vzrdrgpaziba63uhcaro6z23h/spack-src/src/TCProcessor.cpp:672] setting request end: 108861002956407950
```
to be shown in logs.

- similarly for `TRACE_LVLS=-1`:
```
2025-Mar-12 11:52:45,869 DEBUG_10 [dunedaq::opmonlib::opmon::MonitoringTreeInfo dunedaq::opmonlib::MonitorableObject::collect() at /tmp/root/spack-stage/spack-stage-opmonlib-NB_DEV_250312_A9-uged3z5wcepxwu77dq6u7lcifyhzvoqh/spack-src/src/MonitorableObject.cpp:105] Collecting data from local-1x1-config.mlt.def-tc-handler-10000.TCProcessor.def-tc-request-handler
2025-Mar-12 11:52:45,869 DEBUG_1 [void dunedaq::datahandlinglibs::DefaultRequestHandlerModel<ReadoutType, LatencyBufferType>::generate_opmon_data() [with ReadoutType = dunedaq::trigger::TCWrapper; LatencyBufferType = dunedaq::datahandlinglibs::SkipListLatencyBufferModel<dunedaq::trigger::TCWrapper>] at /cvmfs/dunedaq-development.opensciencegrid.org/nightly/NB_DEV_250312_A9/spack-0.22.0/opt/spack/linux-almalinux9-x86_64/gcc-13.2.0/datahandlinglibs-NB_DEV_250312_A9-mce4itn2emdte6cyn47bks3wqjhqlti2/include/datahandlinglibs/models/detail/DefaultRequestHandlerModel.hxx:307] Cleanup request rate: 0 [Hz] Dropped: 0 Occupancy: 0
```

and so on.

bieryAtFnal:
I'm supportive of the spirit of this PR, but I'm confused about the suggested changes.

First, let me say that there is a script in `daqconf` called `daqconf_set_session_env_var` whose purpose is exactly this sort of thing - it sets an env var in the environment of the running DAQ processes.  Before I started looking at the changes in this PR, I started with a fresh software area and ran the following:
```
daqconf_set_session_env_var local-1x1-config config/daqsystemtest/example-configs.data.xml DUNEDAQ_ERS_DEBUG_LEVEL 10
```
This appeared to produce the desired effect.  That is, I saw extra DEBUG messages appear in the logfiles of the DAQ processes.

Comparing the modifications made by that script and the suggested modifications in this PR, I see a couple of differences:

1. the script in `daqconf` makes the following changes:

```
--- a/config/daqsystemtest/example-configs.data.xml
+++ b/config/daqsystemtest/example-configs.data.xml
@@ -63,7 +63,7 @@
 
 <oks-data>
 
-<info name="" type="" num-of-items="6" oks-format="data" oks-version="862f2957270" created-by="gjc" created-on="thinkpad" creation-time="20231207T105549" last-modified-by="eflumerf" last-modified-on="ironvirt9.IRONDOMAIN.local" last-modification-time="20241022T175827"/>
+<info name="" type="" num-of-items="7" oks-format="data" oks-version="862f2957270" created-by="gjc" created-on="thinkpad" creation-time="20231207T105549" last-modified-by="biery" last-modified-on="daq.fnal.gov" last-modification-time="20250313T012522"/>
 
 <include>
  <file path="schema/confmodel/dunedaq.schema.xml"/>
@@ -144,6 +144,7 @@
   <ref class="Variable" id="local-env-ers-warning"/>
   <ref class="Variable" id="local-env-ers-error"/>
   <ref class="Variable" id="local-env-ers-fatal"/>
+  <ref class="Variable" id="session-env-dunedaq-ers-debug-level"/>
  </rel>
  <rel name="disabled">
   <ref class="DFApplication" id="df-02"/>
@@ -178,4 +179,9 @@
  <rel name="opmon_uri" class="OpMonURI" id="local-opmon-uri"/>
 </obj>
 
+<obj class="Variable" id="session-env-dunedaq-ers-debug-level">
+ <attr name="name" type="string" val="DUNEDAQ_ERS_DEBUG_LEVEL"/>
+ <attr name="value" type="string" val="10"/>
+</obj>
+
 </oks-data>

```

I don't see the `<ref class="Variable" id="session-env-dunedaq-ers-debug-level"/>` line in the changes in this PR, and I don't know how the DUNEDAQ_ERS_DEBUG_VALUE gets sent to the DAQ processes without it.  (For reference, I didn't see extra DEBUG messages when I simply used the changes provided in this PR.). Also, it seems rather odd that the `daqconf` script changes entries in example-configs.data.xml, whereas the changes in this PR are in ccm.data.xml.

2. I'm not sure why the TRACE_LVLS env var is set in this PR.  I didn't seem to need it when I ran the test that used `daqconf_set_session_env_var`.  And, I don't know how it gets added to the environment of the running DAQ processes without a line like the extra one mentioned in comment (1).

We can talk over Zoom sometime, if that would help...

MRiganSUSX:
Hi @bieryAtFnal, thank you for looking at this. 

I was unaware of the existence of this script. It does indeed achieve the same thing. Although, I question why then we keep other variables pre-defined, if they all can be set up with this script. In other words, why would `DUNEDAQ_ERS_DEBUG_LEVEL` be treated differently to say `DUNEDAQ_ERS_VERBOSITY_LEVEL`, which is part of the predefined variables. 
I do think the existence of a script that can do this is great, but I also think - if we are aware of a variable, and have others similar variables pre-defined - we should do that here too. 

Regarding 1:
My PR is only adding the variable - creating the object. It does not, by default, add it to any session. However, it makes it much easier to be added if someone wants to, as it will be automatically listed. Some screenshots below:

- Variable objects:
Screenshot from 2025-03-13 11-33-00
Added the `local-env-tlog-debug` for the `DUNEDAQ_ERS_DEBUG_LEVEL`, similar to ie `local-env-ers-verb` for `DUNEDAQ_ERS_VERBOSITY_LEVEL` that is already included.

- default session:
Screenshot from 2025-03-13 11-35-33
I did not change the variables included in example sessions by default.

- adding variable to a session:
Screenshot from 2025-03-13 11-35-53
It is now much simpler to add the variable to a session since it will be given as an option in the dropdown. It takes away the need to memorize the string name (as long as the 'name' of the variable is descriptive enough). 

This line `<ref class="Variable" id="session-env-dunedaq-ers-debug-level"/>` is missing, because as explained above, I did not modify the sessions, just created the objects for these. 
The reason to store in ccm.data.xml was that it hosts other such variables. 

2. the situation is the same for TRACE. I simply made an object so that it is available, but did not add it to any session. I honestly don't really use TRACE and only added it for completeness, since we have this variable and it may be needed by someone. 

Happy to talk about this! Also happy for this to be rejected, it was just something that I had to do when developing and thought it may help others save a minute.

bieryAtFnal:
Thanks for the info @MRiganSUSX ,  that helped.

After thinking about this a little more, I have a minimal suggestion/request and one that goes slightly beyond that.

The minimal suggestion/request has two parts:

1. could we rename the ERS_DEBUG OKS variable to `local-env-dunedaq-ers-debug-level`?  This would allow the `daqconf_set_session_env_var` script to make use of it if/when someone uses that script to set or modify the DUNEDAQ_ERS_DEBUG_LEVEL in a given OKS Session.
2. could we drop the TRACE_LVLS entry?  We typically don't control TRACE messages with env var values.  We set them dynamically with the `tlvls` command.

The hopefully small additional request is the following:

3. could we go ahead and put the ERS_DEBUG variable in the local-1x1-config session?  This would allow users who edit OKS files directly to simply change one number if/when they want to change the DUNEDAQ_ERS_DEBUG_LEVEL.  And, it would serve as an example for how to add a Variable to a different OKS Session.  

Here are the diffs between what I'm proposing and the code that already exists on this branch:
```
(dbt) [biery@daq daqsystemtest]$ git diff .
diff --git a/config/daqsystemtest/ccm.data.xml b/config/daqsystemtest/ccm.data.xml
index c049505..b435ae1 100644
--- a/config/daqsystemtest/ccm.data.xml
+++ b/config/daqsystemtest/ccm.data.xml
@@ -63,7 +63,7 @@
 
 <oks-data>
 
-<info name="" type="" num-of-items="29" oks-format="data" oks-version="862f2957270" created-by="gjc" created-on="thinkpad" creation-time="20231207T110859" last-modified-by="mrigan" last-modified-on="np04-srv-016.cern.ch" last-modification-time="20250312T105152"/>
+<info name="" type="" num-of-items="28" oks-format="data" oks-version="862f2957270" created-by="gjc" created-on="thinkpad" creation-time="20231207T110859" last-modified-by="biery" last-modified-on="daq.fnal.gov" last-modification-time="20250313T155209"/>
 
 <include>
  <file path="schema/confmodel/dunedaq.schema.xml"/>
@@ -222,16 +222,11 @@
  <attr name="value" type="string" val="erstrace,throttle,lstdout"/>
 </obj>
 
-<obj class="Variable" id="local-env-tlog-debug">
+<obj class="Variable" id="local-env-dunedaq-ers-debug-level">
  <attr name="name" type="string" val="DUNEDAQ_ERS_DEBUG_LEVEL"/>
  <attr name="value" type="string" val="0"/>
 </obj>
 
-<obj class="Variable" id="trace-lvl">
- <attr name="name" type="string" val="TRACE_LVLS"/>
- <attr name="value" type="string" val="-1"/>
-</obj>
-
 <obj class="VariableSet" id="ehn1-variables">
  <rel name="contains">
   <ref class="Variable" id="ehn1-env-ers-error"/>
diff --git a/config/daqsystemtest/example-configs.data.xml b/config/daqsystemtest/example-configs.data.xml
index c67a378..550fe25 100644
--- a/config/daqsystemtest/example-configs.data.xml
+++ b/config/daqsystemtest/example-configs.data.xml
@@ -144,6 +144,7 @@
   <ref class="Variable" id="local-env-ers-warning"/>
   <ref class="Variable" id="local-env-ers-error"/>
   <ref class="Variable" id="local-env-ers-fatal"/>
+  <ref class="Variable" id="local-env-dunedaq-ers-debug-level"/>
  </rel>
  <rel name="disabled">
   <ref class="DFApplication" id="df-02"/>
```

MRiganSUSX:
Hi @bieryAtFnal,
1. will do.
2. no issue with that.
3. I wanted to do this as well, but there are some messages that are actually TLOG_DEBUG(0) so adding that does change (increase) what is being printed out by default. I could try with setting it to -1 or similar. What do you think ?

bieryAtFnal:
I ran a test with the ERS_DEBUG level set to -1, and that A) eliminated the extra messages that you mentioned and B) did not cause the system to go crazy.  So, that seems like a reasonable option....

MRiganSUSX:
The latest commits include:
- removing TRACE variable completely
- renaming the variable for `DUNEDAQ_ERS_DEBUG_LEVEL` to `local-env-dunedaq-ers-debug-level`
- adding this variable to the example `local-1x1-config` session
- adding this variable to the set of local variables

I can confirm that setting the default value to -1 results in no debug messages to be shown. 
I've also re-tested that changing this value to any positive integer has the expected effect of showing all debug messages up to (and including) the set level. 

One comment: 
I do not personally like the name `local-env-dunedaq-ers-debug-level` as it is not consistent with the naming for the rest of the variables which do not have the `dunedaq` part of the name, even though the strings they represent have that.

bieryAtFnal:
That suggested naming for the ID of that variable is an artifact of my desire to continue to allow the use of the 'set-session-env-var' script in daqconf to set the ERS DEBUG level, and how that script is set up to work.  It simply uses the name of the environment variable, translated to lower case with hyphens rather than underscores, as the latter part of the variable name.  I can look into using an existing Variable entry in the OKS configuration, if one exists, in the script.  I'll do that.

bieryAtFnal:
I've updated the new Variable ID to the name that I believe is desired.  Hopefully, I've got that correct.

bieryAtFnal (approved):
LGTM, thanks for taking my suggestions into account.

MRiganSUSX:
Thanks for the review.

PR #190, opened 2025-03-17
Modified the fragment_error_bit checks in integtests to skip Trigger fragments

bieryAtFnal:
These changes were made in the spirit of a compromise that allows us to move forward with fragment-error-bit checking, but not have integtests always failing because TriggerCandidate fragments always have the kIncomplete bit set.  

This has been discussed in the 12-Feb Core Software meeting, and it Slack messages in the daq-release-preparation channel on 17-Mar.

I also did a little cleanup in a few integtests, removing stale code.


eflumerf (approved):
Regression tests pass

PR #194, opened 2025-04-10
Boost ASIO socket reader

denizergonul:
A Boost.Asio-based socket reader plugin shall be implemented as a lightweight alternative to the current DPDK-based solution, which is overly complex for low-bandwidth electronic devices.

local-socket-1x1-config (from daqsystemtest/config/daqsystemtest/example-configs.data.xml) is a session configuration with a socket reader application accompanied by a fake socket writer application.

image


config/daqsystemtest/ccm.data.xml
```diff
@@ -136,6 +137,16 @@
  <rel name="fsm" class="FSMconfiguration" id="fsmConf-test"/>
 </obj>
 
+<obj class="RCApplication" id="socket-root-controller">
```
plasorak, line 140: I think you may be able to just use the `root-controller` here.

config/daqsystemtest/example-configs.data.xml
```diff
@@ -100,6 +102,16 @@
...
 </obj>
 
+<obj class="Segment" id="socket-root-segment">
+ <rel name="segments">
+  <ref class="Segment" id="socket-ru-segment"/>
+  <ref class="Segment" id="df-segment"/>
+  <ref class="Segment" id="trg-segment"/>
+  <ref class="Segment" id="hsi-fake-segment"/>
+ </rel>
+ <rel name="controller" class="RCApplication" id="socket-root-controller"/>
```
plasorak, line 112: ... and here.

config/daqsystemtest/ru-segment.data.xml
```diff
@@ -328,6 +346,16 @@
  <rel name="fsm" class="FSMconfiguration" id="FSMconfiguration_noAction"/>
 </obj>
 
+<obj class="RCApplication" id="socket-ru-controller">
```
plasorak, line 349: Also think it may work with the ru-controller here too.

plasorak (commented):
Just a couple of comments in passing.

roland-sipos (approved):
LGTM

roland-sipos:
Please fix based on Pierre's comments.

config/daqsystemtest/example-configs.data.xml
denizergonul: Thank you for pointing this out, but I'm facing this error when I try to make the change:
image

config/daqsystemtest/ru-segment.data.xml
denizergonul: Same error:
image

config/daqsystemtest/example-configs.data.xml
plasorak: Fair enough, in that case you can merge it then.

PR #199, opened 2025-05-13
Updated readout_type_scan.py to use the new DAPHNE data binary file…

bieryAtFnal:
…from run 36012 that doesn't have invalid peaks marked as 'found' like the previous file that we were using.

As part of this, I needed to update some of the expected fragment sizes, since the data in the new file has different characteristics than the old file.

This can be tested by running
```
daqsystemtest_integtest_bundle.sh -f1 -l1
```
after including this branch in a software area that has been set up for v5.3.2 testing (e.g. as described [here](https://github.com/DUNE-DAQ/daqconf/wiki/Setting-up-a-fddaq%E2%80%90v5.3.2-software-area)).

### DIFF
--- a/integtest/readout_type_scan.py
+++ b/integtest/readout_type_scan.py
@@ -55,11 +55,11 @@ daphne_frag_params = {
     "fragment_type_description": "DAPHNE",
     "fragment_type": "DAPHNE",
     "expected_fragment_count": number_of_data_producers,
-    "min_size_bytes": 30000,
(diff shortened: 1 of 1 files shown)


jcfreeman2 (approved):
I've confirmed that while a standard patch v5.3.1-based test build `NFDT_DEV_250514_A9` runs into problems with `readout_type_scan.py`, if I build daqsystemtest on this feature branch and re-run, things are fine. Approved.

PR #205, opened 2025-05-22
Added an option (-k) to the bundle script to allow users to specify a…

bieryAtFnal:
… pipe-delimited list of test names to be run (the pipe-delimited list is passed to egrep).

This allows users to more easily run tests without remembering the index of the test(s) of interest.

Examples included:

- `daqsystemtest_integtest_bundle.sh -k example`
- `daqsystemtest_integtest_bundle.sh -f5 -k 3ru`
- `daqsystemtest_integtest_bundle.sh -k 'min|small'`

I chose the letter 'k' for this option in an attempt to be similar to the "pytest -k" option.  However, since the bundle script is currently written in bash and pytest is written in python, the format of the arguments to the "-k" options are different.

- the argument to the bundle script expects 'egrep'-style search arguments (pipe-delimited)
- the argument to the pytest utility supports Python "or" syntax

### DIFF
--- a/scripts/daqsystemtest_integtest_bundle.sh
+++ b/scripts/daqsystemtest_integtest_bundle.sh
@@ -14,6 +14,7 @@ Options:
     -h, --help : prints out usage information
     -f <zero-based index of the first test to be run, default=0>
     -l <zero-based index of the last test to be run, default=${last_test_index}>
+    -k <pipe-delimited string to select which tests will be run ('egrep -i' match to test name)>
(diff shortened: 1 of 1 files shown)


eflumerf (approved):

eflumerf (approved):
I had to add a workaround for the "no -k" case, since in my shell at least, the `'*'` was simply globbing the files and directories of the current working directory.

bieryAtFnal:
Thanks for catching that and providing a fix!

PR #206, opened 2025-05-22
Several changes to help improve integtest console output: …

bieryAtFnal:
…turned off the pytest traceback; added information about correct number of data files produced (or not); added the flushing of output in the long_window_readout_test so that the listing of files before deletion output isn't garbled.

The current situation in which the `long_window_readout_test` is failing provides a nice opportunity to improve some of our integtest output messages.

1. Now that we have started including the check that the correct number of data files has been produced in our "all_ok" summary status, it can be a little hard to tell whether that individual check has succeeded or passed.  To help in this regard, I've added code to print out messages in either case to the 3 integtests that needed it.  @eflumerf , if you don't like my model of copying the HeavyWhiteCheckMark and PoliceCarsRevolvingLight pattern from integrationtest/data_file_checks.py, we can talk about a different way to help indicate success or failure.
2. To avoid garbled printouts associated with the removal of large data files at the end of the `long_window_readout_test`, I set the default behavior of the Python `print()` command to flush its output in that integtest.
3. Based on a recommendation from Alessandro, I have added control of the `pytest` traceback to the `pytest.ini` file.  This turns off the `pytest` traceback in the case of a test failure since that traceback information is often more confusing than helpful.

To test these changes, one can run the `long_window_readout_test`, `fake_data_producer_test`, and `readout_type_scan` and observe the changes in the output messages.

### DIFF
--- a/integtest/fake_data_producer_test.py
+++ b/integtest/fake_data_producer_test.py
@@ -159,6 +159,11 @@ def test_data_files(run_nanorc):
     # Run some tests on the output data file
     all_ok = True
     all_ok &= len(run_nanorc.data_files) == expected_number_of_data_files
+    print("") # Clear potential dot from pytest
+    if all_ok:
+        print(f"\N{WHITE HEAVY CHECK MARK} The correct number of raw data files was found ({expected_number_of_data_files})")
+    else:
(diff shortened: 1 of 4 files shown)


eflumerf (approved):
LGTM. I think I like keeping consistency of how pass/fail is reported.

PR #213, opened 2025-06-16
Trigger bitwords integtest

MRiganSUSX:
Requires: https://github.com/DUNE-DAQ/integrationtest/pull/108

An integration test for the trigger bitwords. 

Description:
```
Integration Test for Trigger Bitword Configurations
This test module validates DAQ system behavior under various trigger bitword configurations.
It uses multiple configuration variants (e.g., no bitword, prescale, timing, supernova, etc.)
and runs a controlled run control session to verify:
- Successful execution of the run using the run control
- Correct generation of log files and absence of unexpected errors
- Proper creation and content of output data files
- Expected trigger types and trigger multiplicity checks
  (Multiplicity here refers to a Trigger Decision composed of multiple different
   Trigger Candidate types that have been merged)
Detailed case descriptions:
- No bits: A default case with no trigger bitwords configured; all TC types in the config should appear.
- Prescale / Timing bits: A single specific bitword is enabled; only the corresponding TC type should be present.
- Supernova bit: A bitword not expected to participate in the run; no raw data should be produced.
- Series bit: Two independent trigger bitwords enabled; both TC types should appear in the data.
- Coincidence bit: A composite bitword combining two TC types; this triggers merged TDs, resulting in
  multiplicity due to the TC coincidence mechanism.
  ```
  
The configurations blocks are defined in `daqsystemtest/integtest/trigger_bitwords_test.py`
There are two new data files checks in `integrationtest/python/integrationtest/data_file_checks.py`:
- `check_tr_trigger_types`: the main purpose is to check whether a data file contains specific TR types:
  - extracts TR trigger types
  - unpacks extracted trigger types
  - converts these to TC types (as bits)
  - compares the final list of bits to the provided one
- `check_tr_type_multiplicity`: main purpose is to check whether a data file contains TRs that were created from TC merging (a TD formed from multiple (different) overlapping TC types)
  - extracts TR trigger types
  - checks whether these are formed from merging different TC types
There are additional 'helper' functions in `integrationtest/python/integrationtest/data_file_check_utilities.py` that were intentionally made to be very generic to allow reuse in tests with other scope as well.

Note:
There are commented out lines in 'integrationtest/python/integrationtest/data_file_checks.py` that were very useful for me when developing, they clearly show what is happening with the bitwords. I left these lines in if ppl find them useful when testing.

### DIFF
--- a/config/daqsystemtest/moduleconfs.data.xml
+++ b/config/daqsystemtest/moduleconfs.data.xml
@@ -63,7 +63,7 @@
 
 <oks-data>
 
-<info name="" type="" num-of-items="78" oks-format="data" oks-version="862f2957270" created-by="gjc" created-on="thinkpad" creation-time="20231207T105629" last-modified-by="mrigan" last-modified-on="np04-srv-031.cern.ch" last-modification-time="20250423T102725"/>
+<info name="" type="" num-of-items="80" oks-format="data" oks-version="862f2957270" created-by="gjc" created-on="thinkpad" creation-time="20231207T105629" last-modified-by="mrigan" last-modified-on="np04-srv-031.cern.ch" last-modification-time="20250423T102725"/>
 
 <include>
  <file path="schema/confmodel/dunedaq.schema.xml"/>
@@ -684,6 +684,18 @@
  <rel name="tp_streams">
   <ref class="TPStreamConf" id="def-tp-stream-1"/>
  </rel>
+</obj>
+
+<obj class="TriggerBitword" id="test-bitword">
+ <attr name="bitword" type="string">
+  <data val="kPrescale"/>
+ </attr>
+</obj>
+
+<obj class="TriggerBitword" id="test-bitword2">
+ <attr name="bitword" type="string">
+  <data val="kTiming"/>
+ </attr>
 </obj>
 
 <obj class="SamplesOverThresholdMinima" id="def-sot-minima">
--- a/integtest/example_system_test.py
+++ b/integtest/example_system_test.py
@@ -62,16 +62,10 @@ hsi_frag_params = {
                               "default": {"min_size_bytes":  72, "max_size_bytes":  72} }
 }
 ignored_logfile_problems = {
-    "-controller": [
-        "Worker with pid \\d+ was terminated due to signal",
-        "Connection '.*' not found on the application registry",
-    ],
     "local-connection-server": [
         "errorlog: -",
-        "Worker with pid \\d+ was terminated due to signal",
         r"Worker \(pid:\d+\) was sent SIGHUP"
-    ],
-#    "log_.*": ["connect: Connection refused", "Connection reset by peer", "end of stream"],
+    ]
 }
 
 # The arguments to pass to the config generator, excluding the json
@@ -130,6 +124,13 @@ nanorc_command_list = (
 
 def test_nanorc_success(run_nanorc):
     current_test = os.environ.get("PYTEST_CURRENT_TEST")
+    match_obj = re.search(r".*\[(.+)-run_nanorc0\].*", current_test)
+    if match_obj:
+        current_test = match_obj.group(1)
+    banner_line = re.sub(".", "=", current_test)
+    print(banner_line)
+    print(current_test)
+    print(banner_line)
 
     if not host_is_at_ehn1(hostname) and "EHN1" in current_test:
         pytest.skip(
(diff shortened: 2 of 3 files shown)


bieryAtFnal:
This new test looks great.  

A few questions and suggestions:

1. Is the `daqsystemtest` repo the best place for this integration/regression test?  Another option might be the `trigger` repo, which already has some integtests that cover functionality that is somewhat specific to the trigger.  Maybe we can/should talk about this at an upcoming meeting, and apologies if this has already been discussed.
2. Could a sentence be added to the nice introduction at the top of the test file to remind readers that trigger bitwords are used to control which triggers are allowed to be promoted to trigger decisions?  And, that this functionality is independent of whatever trigger sources are included in the system configuration that is being used?
3. It looks like the logic in the `integrationtest/data_file_check_utilities::check_multi_TR_type()` function returns False if *all* TR types that are passed in are single-bit values and True if one or more TR types that are passed in have multiple bits set.  So, this function can't be used to require that all specified TR types have multiple bits set.  However, it appears that the "Coincidence bit" test-configuration is providing exactly that - a data sample in which each Trigger Record does result from the coincidence of multiple triggers.  Is it important to verify that *all* TRs have multiple bits set in the "Coincidence bit" configuration?
4. It seems that several of the test-configurations could accidentally produce a Trigger Record with multiple bits set, and that would cause the `check_multi_TR_type()` data-check to fail.  Would it make sense to set the `merge_overlapping_tcs` configuration parameter to False for some of the test-configurations to reduce the chance of this accidentally happening?
5. I couldn't seem to select just the "No bits" test-configuration with the `pytest -k No` option.  Is there a recommended way to do that?
6. There a couple of changes that I'd like to suggest that can/should also be made to the `example_system_test` (which I believe was used as an example for this new test).  Those are the following"
    * print out the name of the test-configuration immediately before the results of the data-checks
    * remove a few obsolete ignored phrases
    * I can provide sample code changes that show exactly what I mean, if that would be helpful

Thanks!

MRiganSUSX:
Hi @bieryAtFnal,
Thank you for your comments. My thoughts:

1. This has been discussed at the SWIT meeting. I don't have a personal preference, but my suggestion would be to make these tests available widely. Otherwise, we are open to developers outside of the trigger breaking the trigger code (such as replay). I hoped adding it to daqsystemtest would make it easy to be added to the `daqsystemtest_integtest_bundle.sh`, hopefully increasing the likelihood of catching breaking changes early. 
2. Added.
3. This is definitely a good line of thinking. I think, especially implementing your suggestion from 4. there is no need to change the function as it fits the purpose of this test. But I am happy to change it if you think it is necessary. 
[as a sidenote, it was actually quite hard to get multirecords, and I suspect there is something wrong with it - but not as a result of this test of course]
4. Great idea.  Added.
5. Ah. This was quite a pain, surprisingly. With my previous choice of names, one would have to do something like
`pytest --collect-only trigger_bitwords_test.py -k "No and bits"`
to get no bits. I have now changed the names so it is much simpler.
6. Please do provide the sample changes. I'm happy to work on that.
Thank you.

bieryAtFnal:
@MRiganSUSX , thanks for your reply.

Continuing the conversation...

1. Yes, at the moment, I'm leaning toward leaving this test in the `daqsystemtest` repo until we have a good picture of how to ensure that tests get run often enough.  Let's keep thinking and talking about this.
2. Thanks for adding the extra information in the introductory text in the test.
3. Leaving the `check_multi_TR_type()` function as it is, is fine with me.  If we come up with a use case that would benefit from a change, we can make the change then.  And, I should have said that A) I can definitely imagine using some of the new functions that you created in other integtests that I've worked on, so I appreciate having them available, and B) I also have occasionally seen unexpected behavior when TCs overlap, but haven't followed up on that yet.
4. I'm glad to hear that you also liked the idea of setting  `merge_overlapping_tcs` to False in certain cases.
5. Thanks for changing the config names to make the use of the "-k" option easier.
6. It was easiest to demonstrate the minor mods that I mentioned by simply modifying (and testing) the code.  The candidate changes are on the `kbiery/tweaks_to_bitwords_and_example_tests` branch, and I hope they seem reasonable.

MRiganSUSX:
@bieryAtFnal thank you for your changes, I've merged your branch to this one (hopefully that is ok).

Otherwise, I think all comments were addressed both here and in https://github.com/DUNE-DAQ/integrationtest/pull/108.
Please review/approve at your convenience.

bieryAtFnal (approved):
Looks good to me.  Thanks!

PR #215, opened 2025-06-24
Updates for PDS/TPCRawDataProcessor Integration

ramarine:
PDS/TPCRawDataProcessor included in configs. With this version all into tests pass.

One thing worth checking is if  in `moduleconfs.data.xml`, for the` def-pds-stream-link-handler`  do I also have to set the TPCRawDataProcessor to a `PDSRawDataProcessor`?

### DIFF
--- a/config/daqsystemtest/moduleconfs.data.xml
+++ b/config/daqsystemtest/moduleconfs.data.xml
@@ -63,7 +63,7 @@
 
 <oks-data>
 
-<info name="" type="" num-of-items="84" oks-format="data" oks-version="862f2957270" created-by="gjc" created-on="thinkpad" creation-time="20231207T105629" last-modified-by="ramarine" last-modified-on="np04-srv-017.cern.ch" last-modification-time="20250623T201605"/>
+<info name="" type="" num-of-items="82" oks-format="data" oks-version="862f2957270" created-by="gjc" created-on="thinkpad" creation-time="20231207T105629" last-modified-by="ramarine" last-modified-on="np04-srv-017.cern.ch" last-modification-time="20250625T100027"/>
 
 <include>
  <file path="schema/confmodel/dunedaq.schema.xml"/>
@@ -83,6 +83,8 @@
  <comment creation-time="20250530T152100" created-by="dergonul" created-on="np04-srv-015.cern.ch" author="dergonul" text="Added CRT action plans"/>
  <comment creation-time="20250530T152404" created-by="dergonul" created-on="np04-srv-015.cern.ch" author="dergonul" text="Added CRT action plans"/>
  <comment creation-time="20250623T201605" created-by="ramarine" created-on="np04-srv-017.cern.ch" author="ramarine" text="PDSlinked"/>
+ <comment creation-time="20250625T095129" created-by="ramarine" created-on="np04-srv-017.cern.ch" author="ramarine" text="deleted latency buffer"/>
+ <comment creation-time="20250625T100027" created-by="ramarine" created-on="np04-srv-017.cern.ch" author="ramarine" text="def-pds-tp-handler deleted"/>
 </comments>
 
 
@@ -337,7 +339,7 @@
  <attr name="generate_timesync" type="bool" val="1"/>
  <rel name="request_handler" class="RequestHandler" id="def-data-request-handler"/>
  <rel name="latency_buffer" class="LatencyBuffer" id="def-latency-buf"/>
- <rel name="data_processor" class="PDSRawDataProcessor" id="PDSRawDataProc"/>
+ <rel name="data_processor" class="PDSRawDataProcessor" id="def-pds-processor"/>
 </obj>
 
 <obj class="DataHandlerConf" id="def-pds-stream-link-handler">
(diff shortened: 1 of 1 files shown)


config/daqsystemtest/moduleconfs.data.xml
```diff
@@ -571,6 +589,14 @@
  <attr name="expected_timestamp_step" type="s32" val="1000"/>
 </obj>
 
+<obj class="PDSRawDataProcessor" id="PDSRawDataProc">
```
wesketchum: probably should follow naming conventions for other ids. something like `def-pds-processor`

config/daqsystemtest/moduleconfs.data.xml
```diff
@@ -559,6 +569,14 @@
  <attr name="preallocation" type="bool" val="0"/>
 </obj>
 
+<obj class="LatencyBuffer" id="tp-latency-buf-numa0">
```
wesketchum: why is this object being created in this PR? Is this something? Is this something to do with the new PDS/TPCRawDataProcessor configs?

config/daqsystemtest/moduleconfs.data.xml
```diff
@@ -345,7 +346,16 @@
  <attr name="generate_timesync" type="bool" val="1"/>
  <rel name="request_handler" class="RequestHandler" id="def-data-request-handler"/>
  <rel name="latency_buffer" class="LatencyBuffer" id="def-latency-buf"/>
- <rel name="data_processor" class="RawDataProcessor" id="def-wib-processor"/>
+ <rel name="data_processor" class="TPCRawDataProcessor" id="def-wib-processor"/>
+</obj>
+
+<obj class="DataHandlerConf" id="def-pds-tp-handler">
```
wesketchum: why is there a different data handler conf for PDS TPs vs. TPC TPs? Shouldn't the PDS TPs still use the `def-tp-handler`?

wesketchum (changes requested):

config/daqsystemtest/moduleconfs.data.xml
```diff
@@ -571,6 +589,14 @@
  <attr name="expected_timestamp_step" type="s32" val="1000"/>
 </obj>
 
+<obj class="PDSRawDataProcessor" id="PDSRawDataProc">
+ <attr name="queue_sizes" type="u32" val="10000"/>
+ <attr name="thread_names_prefix" type="string" val="postproc-"/>
+ <attr name="latency_monitoring" type="bool" val="1"/>
+ <attr name="channel_map" type="string" val="DefaultPDSChannelMap"/>
+ <attr name="rawdataprocessor_id" type="u32" val="234"/>
```
wesketchum: remember to remove this here as well

wesketchum:
Also @ramarine , I don't think any of these new objects are getting used in tests anyway. We will also need to make modifications to things like here so they get picked up.

https://github.com/DUNE-DAQ/daqsystemtest/blob/f0b3db8b02b5b5853326d6a5fe5b52aef346729a/integtest/readout_type_scan.py#L184

config/daqsystemtest/moduleconfs.data.xml
```diff
@@ -571,6 +589,14 @@
  <attr name="expected_timestamp_step" type="s32" val="1000"/>
 </obj>
 
+<obj class="PDSRawDataProcessor" id="PDSRawDataProc">
+ <attr name="queue_sizes" type="u32" val="10000"/>
+ <attr name="thread_names_prefix" type="string" val="postproc-"/>
+ <attr name="latency_monitoring" type="bool" val="1"/>
+ <attr name="channel_map" type="string" val="DefaultPDSChannelMap"/>
```
alessandrothea: What is a "DefaultPDSChannelMap"?

config/daqsystemtest/moduleconfs.data.xml
```diff
@@ -372,7 +382,7 @@
  <attr name="generate_timesync" type="bool" val="1"/>
  <rel name="request_handler" class="RequestHandler" id="def-data-request-handler"/>
  <rel name="latency_buffer" class="LatencyBuffer" id="def-latency-buf"/>
- <rel name="data_processor" class="RawDataProcessor" id="def-wib-processor"/>
+ <rel name="data_processor" class="TPCRawDataProcessor" id="def-wib-processor"/>
```
alessandrothea, line 378: Probably unrelated to this PR but still: 
Why does the `def-ta-handler` have a relationship to a `TPC/RawDataProcessor`?
It doesn't process raw data/

config/daqsystemtest/moduleconfs.data.xml
ramarine: That is the plugin where the mapping of the pds is implemented. If you remember you started with a functional example of how this is implemented and then Wes modified it to have a first working version.
 It would be in `/detchannelmaps/plugins/DefaultPDSChannelMap`

config/daqsystemtest/moduleconfs.data.xml
ramarine: renamed it to def-pds-processor

config/daqsystemtest/moduleconfs.data.xml
ramarine: true.  def-ads-tp-handler removed.

config/daqsystemtest/moduleconfs.data.xml
ramarine: the relationship with RawDataProcessor is indeed not implemented in this PR. Maybe we should ask on slack.

config/daqsystemtest/moduleconfs.data.xml
wesketchum: Correct. This is the 'map' that defines the offline chan as slot*100+daphne_channel, replicating the previous functionality. Will be working with PDS experts to get a new correct-with-offline map in place in the future.

wesketchum (approved):

PR #217, opened 2025-06-30
TPReplay integtest

MRiganSUSX:
Requires: https://github.com/DUNE-DAQ/integrationtest/pull/111

An integration test for the TPReplay application. 

Description:
```
Integration test configuration and test suite for TPReplay in the DAQ system.
This script sets up temporary configurations for two environments:
    - np02 tpreplay
    - np04 tpreplay
It does the following:
1. Creates a temporary configuration DB file (via OKS) for each session.
2. Populates the config DB with TPStream and SourceID objects.
3. Customizes runtime configuration through deep config substitutions.
4. Runs a pre-defined nanorc command sequence (boot → start → stop → terminate).
5. Validates:
    - Nanorc command success
    - Presence and correctness of log files
    - Data file contents (number of SIDs, file count)
Tests are structured using `pytest` and use fixtures provided via 
`integrationtest.integrationtest_drunc`.
Temporary config directories are cleaned up using `atexit` once the test completes.
```

The configurations blocks are defined in `daqsystemtest/integtest/tpreplay_test.py`
There is one new data files check in `integrationtest/python/integrationtest/data_file_checks.py`:
- `check_n_unique_sids`: purpose is to check whether the data file contains data from all expected sourceIDs (X TP buffers, 1 TA buffer, 1 TC buffer):
  - gets records from HDF5
  - loops over records, extracts all unique sourceIDs per fragment type (TP, TA, TC)
  - compares against expected values for these
  - to get the number of unique SourceIDs, an additional helper function was added to `integrationtest/python/integrationtest/data_file_check_utilities.py`, `sid_key`: which generates a unique key tuple for a Source ID object based on its subsystem and id.
  
These configurations need to create some objects depending on the input data. For a 'normal' replay, this is handled by a python script. Here, the approach is to copy the full configuration to a temp directory and modify it there before pytest takes over. 
The input data files are loaded as asset files.

### DIFF
--- a/integtest/tpreplay_test.py
+++ b/integtest/tpreplay_test.py
@@ -1,0 +1,189 @@
+import copy
+import conffwk
+import os
+import pathlib
+import pytest
+import random
+import re
+import string
+
+import integrationtest.data_classes as data_classes
+import integrationtest.data_file_checks as data_file_checks
+import integrationtest.log_file_checks as log_file_checks
+
+from daqconf.consolidate import copy_configuration
+from pathlib import Path
+
+pytest_plugins = "integrationtest.integrationtest_drunc"
+
+# Run setup
+run_duration = 15  # seconds
+check_for_logfile_errors = True
+ignored_logfile_problems = {
+    "-controller": [
+        "Worker with pid \\d+ was terminated due to signal",
+        "Connection '.*' not found on the application registry",
+    ],
+    "local-connection-server": [
+        "errorlog: -",
+        "Worker with pid \\d+ was terminated due to signal",
+        r"Worker \(pid:\d+\) was sent SIGHUP"
+    ],
+    "config_mlt": [
+        "Trigger is inhibited"
+    ],
+    "config_dfo": [
+        "that was busy with"
+    ]
+#    "log_.*": ["connect: Connection refused", "Connection reset by peer", "end of stream"],
+}
+
+### Config setup
+# Create temp config
+temp_path = "tmp_config"
+path = (Path(os.getcwd()) / temp_path).resolve()
+path.mkdir(parents=True, exist_ok=True)
+copy_configuration(path, [os.path.dirname(__file__) + "/../config/daqsystemtest/example-configs.data.xml"])
+local_db = conffwk.Configuration("oksconflibs:" + temp_path + "/example-configs.data.xml")
+
+common_config_obj = data_classes.drunc_config()
+common_config_obj.op_env = "test"
+common_config_obj.tpg_enabled = False
+common_config_obj.config_db = (
+        os.path.dirname(__file__) + "/" + temp_path + "/example-configs.data.xml"
+)
+
+# Get default tpreplay config
+tpreplay_local_conf = copy.deepcopy(common_config_obj)
+tpreplay_local_conf.session = "local-tpreplay-config"
+
+# Get necessary dal objects
+a_source_id_dal = local_db.get_dal(class_name="SourceIDConf", uid="tpreplay-tp-srcid-100000")
+a_tpstream_conf_dal = local_db.get_dal(class_name="TPStreamConf", uid="def-tp-stream-1")
+
+## setup TPStream files
+first_tpstream_file = copy.deepcopy(a_tpstream_conf_dal)
+second_tpstream_file = copy.deepcopy(a_tpstream_conf_dal)
+first_tpstream_file.id = "tp-stream-1"
+first_tpstream_file.filename = "/nfs/home/mrigan/data/np02vd_tp_run036721_0100_tp-stream-writer_tpw_0_20250603T134642.hdf5"
+first_tpstream_file.index = 1
+second_tpstream_file.id = "tp-stream-2"
+second_tpstream_file.filename = "/nfs/home/mrigan/data/np02vd_tp_run036721_0101_tp-stream-writer_tpw_0_20250603T134859.hdf5"
+second_tpstream_file.index = 2
+local_db.update_dal(first_tpstream_file)
+local_db.update_dal(second_tpstream_file)
+
+## setup SourceIDs
+all_sourceIDs = []
+for a_sid_counter in range(1, 8):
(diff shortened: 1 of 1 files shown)


ArturSztuc (approved):
Tested and works, LGTM

bieryAtFnal:
Is it expected that the NP04 part of the test only creates 2 TriggerRecords in a 20-second run, and 2 of the 3 TriggerPrimitive fragments are empty?

In the NP02 part of the test, is it expected that only 1 of the 2 TP fragments is not empty?

bieryAtFnal:
Also, I'm a little surprised that there aren't TPStream files generated...

MRiganSUSX:
Hi @bieryAtFnal, 
Thank you for your comments.
Answers:

1. NP04 only seeing 2 TRs
This is indeed what I see running replay 'offline' (outside of integtest but with the same configuration and drunc commands): 
Screenshot from 2025-07-30 15-58-50
Screenshot from 2025-07-30 15-59-35

There are several reasons for this:
- there is filtering applied for both induction planes, so very few planes contribute
- the actual TP rates for the planes that contribute is realistically just low (see above)
- the integtest is using default processors from the config, which in this case are Prescales (both TAM and TCM) with configs 1000:100. 

2. Usually all but 1 fragment is empty (both np02/np04)
This is again the results of above mentioned reasons. We are using Prescale algo which just promotes whatever TP was 'lucky'. There are only very few contributing planes (3 in the case of np04 and only 2 in the case of np02) and the rates are low, so the probability of TPs overlapping in the small readout window is low. 

This is also what I see 'offline':
```
HDF5LIBS_TestDumpRecord test_raw_run000101_0000_df-01_dw_0_20250730T134333.hdf5 -n 3
2025-Jul-30 15:46:32,484 LOG [main(...) at /tmp/root/spack-stage/spack-stage-hdf5libs-NB_DEV_250730_A9-blbtppwe2eta7nwvhdjcvcozmsdpiaqz/spack-src/test/apps/HDF5LIBS_TestDumpRecord.cpp:109] 
File name: test_raw_run000101_0000_df-01_dw_0_20250730T134333.hdf5
	Recorded size: 54280
	Uncompressed raw data size: 54280
Record type = TriggerRecord
2025-Jul-30 15:46:32,516 LOG [main(...) at /tmp/root/spack-stage/spack-stage-hdf5libs-NB_DEV_250730_A9-blbtppwe2eta7nwvhdjcvcozmsdpiaqz/spack-src/test/apps/HDF5LIBS_TestDumpRecord.cpp:130] 
	Run number: 101
	File index: 0
	Creation timestamp: 1753883013090
	Writer app name: df-01_dw_0
	Compression level: 0
2025-Jul-30 15:46:32,516 LOG [main(...) at /tmp/root/spack-stage/spack-stage-hdf5libs-NB_DEV_250730_A9-blbtppwe2eta7nwvhdjcvcozmsdpiaqz/spack-src/test/apps/HDF5LIBS_TestDumpRecord.cpp:146] 
Number of records: 53
	First record: 1,0
	Last record: 53,0
2025-Jul-30 15:46:32,517 LOG [main(...) at /tmp/root/spack-stage/spack-stage-hdf5libs-NB_DEV_250730_A9-blbtppwe2eta7nwvhdjcvcozmsdpiaqz/spack-src/test/apps/HDF5LIBS_TestDumpRecord.cpp:189] 
	TriggerRecordHeader: check_word: 33334444, version: 4, trigger_number: 1, run_number: 101, trigger_timestamp: 109465594624868011, trigger_type: 32, error_bits: 0, num_requested_components: 4, sequence_number: 0, max_sequence_number: 0, element_id: { subsystem: TR_Builder id: 1 }
	Trigger_Primitive fragment with SourceID Trigger_0x00000001 from subdetector DAQ has size = 144 -----
		Readout window begin = 0, end = 64 (relative to the trigger_timestamp)
		Number of TPs in this fragment=3, size of TP data structure=24, size of Fragment Header=72
		First TP flag = 0, TP detid = 10
		First TP start time=109465594624868011, samples to peak=0, and samples over threshold=2
	Trigger_Primitive fragment with SourceID Trigger_0x00000002 from subdetector DAQ has size = 72 -----
		Readout window begin = 0, end = 64 (relative to the trigger_timestamp)
		*** Empty fragment! Moving to next fragment. ***
	Trigger_Activity fragment with SourceID Trigger_0x0000044c from subdetector DAQ has size = 184 -----
		Readout window begin = 0, end = 64 (relative to the trigger_timestamp)
		Number of TAs in this fragment=1, overall number of referenced TPs=1, size of TA data=88
		First TA type = 1, TA algorithm = 2, number of TPs = 1
		First TA start time=109465594624868011, end time=109465594624868075, and activity time=0
	Trigger_Candidate fragment with SourceID Trigger_0x00002710 from subdetector DAQ has size = 208 -----
		Readout window begin = 0, end = 64 (relative to the trigger_timestamp)
		Number of TCs in this fragment=1, overall number of referenced TAs=1, size of TC data=56
		First TC type = kPrescale (5), TC algorithm = 3, number of TAs = 1
		First TC start time=109465594624868011, end time=109465594624868075, and candidate time=109465594624868011
```

3. No tpstream files generated
The test is using modified copy of the `local-tpreplay-config` which has `tp-stream-writer` disabled. 

Summary:
Everything you observe is expected (and the same result is obtained using replay 'offline' - outside of integtest). They are all consequences of the configuration (filtering, choice and setup of algos) and the actual properties of data files (low rates). The files were not chosen to be 'great' source of TPs, just anything that can replicate the system. The test itself is not meant to test data-quality (at least not for now) but rather to just confirm whether replaying (using the TPReplay application) is possible with some stable files.

bieryAtFnal (approved):
Thanks for the replies.  I now understand that this test is mainly focused on the mechanics of  running the TPReplay tools, not checking the quality of the data produced.  That's fine, but it seems to me that the latter would be useful someday.  Once things settle down (i.e. post-v5.4.0), I'd like to talk about moving this test to the `trigger` repo.

Issue #218, opened 2025-06-30
The sample CRT configs in example-configs.data.xml fail to successfully complete the stop-trigger-sources transition


bieryAtFnal:
These tests were run with fddaq-v5.3.2-rc4.

When I ran each of the following commands:

```
drunc-unified-shell ssh-standalone config/daqsystemtest/example-configs.data.xml local-crt-bern-1x1-config ${USER}-local-test boot wait 5 conf wait 3 start --run-number 111 enable-triggers wait 10 disable-triggers drain-dataflow stop-trigger-sources stop scrap terminate
```

or

```
drunc-unified-shell ssh-standalone config/daqsystemtest/example-configs.data.xml local-crt-grenoble-1x1-config ${USER}-local-test boot wait 5 conf wait 3 start --run-number 121 enable-triggers wait 10 disable-triggers drain-dataflow stop-trigger-sources stop scrap terminate
```

the system seemed to time out while trying to complete the `stop-trigger-sources` transition.  

I noticed the following error messages in the logs (although these may have been symptoms of the problem rather than causes):

```
$ egrep 'ERROR|WARN' log_*.txt
log_biery_biery-local-test_crt-grenoble-root-controller.txt:[2025/06/30 12:12:44] ERROR      controller.py:510              controller:                                   Something wrong happened while sending the command to crt-grenoble-controller: Error raised: Socket closed
log_biery_biery-local-test_crt-grenoble-root-controller.txt:[2025/06/30 12:12:44] ERROR      controller.py:513              controller:                                   Socket closed
log_biery_biery-local-test_crt-grenoble-root-controller.txt:[2025/06/30 12:12:44] ERROR      controller.py:532              controller:                                   Failed to propagate execute_fsm_command to crt-grenoble-controller (crt-grenoble-controller) EXCEPTION THROWN: Socket closed
```

And, I noticed that the HDF5 raw data file did not successfully get closed in these tests (it still had the ".writing" suffix).

wesketchum:
I can't manage to reproduce this ...

I'm running on np04-srv-015. I've tried both with a fddaq-v5.3.2-rc4 dev area, and with the release, and in both cases with multiple runs I don't see this happen: no errors or warnings in the logs, hdf5 files closed properly, CRT logs indicating the run_produce() function stopped well ...

@denizergonul , perhaps you're able to try and take a look?

[...]

PR #219, opened 2025-07-02
cleanup in action plan, as recommended by Kurt

wesketchum:
This is in followup to #218 . Still having trouble recreating that error, but this cleans up some of the logs of the systemtest regardless, so is good to do.


wesketchum:
See https://github.com/DUNE-DAQ/crtmodules/pull/10 for testing details.

denizergonul (approved):
Thanks!

PR #226, opened 2025-07-22
Set the tp-handler-group as optional, as TP handlers may not be present when TPG is not used

eflumerf:
Follows https://github.com/DUNE-DAQ/confmodel/pull/76 and https://github.com/DUNE-DAQ/appfwk/pull/346


mroda88 (approved):
LGTM

eflumerf:
appfwk 346 has been merged into its parent PR:
https://github.com/DUNE-DAQ/appfwk/pull/342

PR #228, opened 2025-07-23
Modified the minimal_system_quick_test so that it prints out a message…

bieryAtFnal:
…indicating whether the expected number of data files was found.

I found this useful recently when I was testing changes that temporarily broke this test.  Also, we have made a similar improvement in other integtests, so it seems reasonable to make it here.


jcfreeman2 (approved):
This change seems reasonable, analogous to what happens in `fake_data_producer_test.py`, though not all tests (e.g. `3ru_1df_multirun_test.py`). Intentionally setting an incorrect expected number of files (`7`), it seems like the difference this makes is that rather than crashing out immediately if it doesn't see the expected number of files, it'll still give the per-file report (e.g., `Record count 21 is within a tolerance of 2 from an expected value of 20`, etc.). It may be worth considering if we want to make this standard across all integration tests.

bieryAtFnal:
Thanks, and yes, we should roll this change out to other regression tests as time permits.

PR #229, opened 2025-07-24
Removed the unneeded socket-write-start and socket-write-stop Action …

bieryAtFnal:
…Plans from moduleconfs.data.xml.

These ActionPlans aren't used anywhere in our example configuration(s), so I'd like us to remove them in an effort to help reduce confusing aspects of our sample configs.


denizergonul (approved):
Thanks!

PR #230, opened 2025-07-24
Removed unneeded parts of CRTReaderApplication declarations…

bieryAtFnal:
… (in ru-segment.data.xml )

My understanding is the CRTReaderApplications are simple apps that send data to Readout Apps (that are configured to receive data from them).  As such, I don't believe that CRTReaderApplications have many of the features that are typical of Readout Apps.  

This PR attempts to remove the unneeded parts of the CRTReaderApplication declarations in our example configuration.

It would be great for reviewers to check that these changes are acceptable.  But in any case, we definitely need to clean up these entries, even if I've removed too much here.

I've verified that usual regression tests pass, both ones with and without CRT.


denizergonul (approved):
The reason I kept them in the first place was because I wasn't sure if they're required or not. Now it's more clear, thank you.

PR #231, opened 2025-07-29
Modified entries in ru-segment.data.xml so that the CRT socket-based apps don't reference anything about WIBs...

bieryAtFnal:
...or Hermes in their configuration.

Also, removed some unused Socket-app-related entities.

### DIFF
--- a/config/daqsystemtest/ru-segment.data.xml
+++ b/config/daqsystemtest/ru-segment.data.xml
@@ -351,15 +351,6 @@
  <rel name="net_receiver" class="DPDKReceiver" id="ru-2-rcv"/>
 </obj>
 
-<!--
-<obj class="NetworkDetectorToDaqConnection" id="socket_conn_apa1">
- <rel name="net_senders">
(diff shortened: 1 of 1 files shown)


config/daqsystemtest/ru-segment.data.xml
```diff
@@ -703,9 +662,10 @@
  <rel name="streams">
   <ref class="DetectorStream" id="stream_1001"/>
  </rel>
- <rel name="uses" class="NetworkInterface" id="wib_101_hermes0"/>
+ <rel name="uses" class="NetworkInterface" id="crt_101_asio0"/>
 </obj>
 
+<!--
```
denizergonul: Can we just remove this?

denizergonul (approved):
Thank you.

config/daqsystemtest/ru-segment.data.xml
bieryAtFnal: yes, done

PR #232, opened 2025-07-29
Switched to using a default base data-processor config for DataHandlers th…

bieryAtFnal:
…at don't have a data processor.


roland-sipos (approved):
LGTM,
@denizergonul please test and review the other PR:
https://github.com/DUNE-DAQ/daqsystemtest/pull/231

PR #235, opened 2025-08-11
Emu CRT configurations

denizergonul:
Emu CRT confs for `asiolibs` integtest.

image

The following is the list of relevant branches (where `SkipListLatencyBufferModel` is used as the buffer type):
asiolibs: dte/asio_integtest
daqsystemtest: dte/emu_crt
appmodel: dte/fakereader_crt
fdreadoutlibs: dte/queue_for_crt
fdreadoutmodules: dte/queue_for_crt
crtmodules: dte/crt_integtest


bieryAtFnal (approved):
I have questions about why so many duplicate entries were needed as part of adding the emulated-data CRT sessions, and we'll talk about those later.  Approving what we have, for now.

PR #240, opened 2025-09-02
Reduced sample configuration duplication in example configs

bieryAtFnal:
## Description

I've modified ccm.data.xml, example-configs.data.xml, moduleconfs.data.xml, and ru-segment.data.xml so that instead of having four sets of configuration objects for CRT Bern vs. Grenoble, and FDFakeReaderModule vs. CRTxyzReaderModule, there is only one set of config objects that gets appropriately modified by the integtests that make use of the local-socket-1x1 system configuration.

This was motivated by a desire to reduce complexity of what we see in the example configs.

## Testing Suggestions

Please see the suggestions in the parent Issue, DUNE-DAQ/daq-deliverables#181

### DIFF
--- a/config/daqsystemtest/ccm.data.xml
+++ b/config/daqsystemtest/ccm.data.xml
@@ -129,46 +129,6 @@
  <attr name="type" type="enum" val="file"/>
 </obj>
 
-<obj class="RCApplication" id="crt-bern-root-controller">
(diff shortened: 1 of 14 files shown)


eflumerf (approved):
I ran `dbt-build --integtest`, and things functioned normally

PR #244, opened 2025-10-02
[CCM] K8s: changing local con port

MRiganSUSX:
To make local connectivity service work with K8s we need to use ports in K8s allowed range for the nodeport service: https://kubernetes.io/docs/concepts/services-networking/service/#type-nodeport

This solution makes it possible to use either of the existing process managers. 

More details on the k8s side: https://github.com/DUNE-DAQ/drunc/pull/603.
Associated daqconf PR: https://github.com/DUNE-DAQ/daqconf/pull/603

### DIFF
--- a/config/daqsystemtest/ccm.data.xml
+++ b/config/daqsystemtest/ccm.data.xml
@@ -87,7 +87,7 @@
  <attr name="commandline_parameters" type="string">
   <data val="Process PID $$;"/>
   <data val="trap &apos;pkill -INT -P $$&apos; EXIT;"/>
-  <data val="gunicorn -b 0.0.0.0:30005 --workers=1 --worker-class=gthread --threads=2 --timeout 5000000000 --log-level=info connection-service.connection-flask:app"/>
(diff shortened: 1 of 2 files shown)


bieryAtFnal:
Hi,
The switch to using the CONNECTION_PORT variable broke the `daqconf_set_connectivity_service_port` script in `daqconf`.  This script is used to set a unique port for users running tests on the same computer, so we want to keep it functional.

MRiganSUSX:
There is now an additional PR in daqconf to fix this script: https://github.com/DUNE-DAQ/daqconf/pull/603.

MRiganSUSX:
Should mention:
Tested both PRs by:
- running the script manually:
```
daqconf_set_connectivity_service_port local-1x1-config config/daqsystemtest/example-configs.data.xml
```
and looking at the corresponding xml files
- running the example_system_test.py
- running the whole daqsystemtest_integtest_bundle.sh
- running both offline and online custom tests.

MRiganSUSX:
Superseded by https://github.com/DUNE-DAQ/daqsystemtest/pull/260.

PR #247, opened 2025-10-09
SNB Transform demonstrator

eflumerf:
Part of: https://github.com/DUNE-DAQ/daq-deliverables/issues/186

Adds ActionPlans for readout-start and readout-stop for the case where the SNB Transform input source is used.


andrewmogan (approved):
LGTM.

See comments in [daq-deliverables #189](https://github.com/DUNE-DAQ/daq-deliverables/issues/186) for tests performed.

PR #248, opened 2025-10-15
Moving processor type definition in integration test configurations

xinyue-uoft:
# Current issue

## Test and observation

If I run the `tpstream_writing_test.py` with the current up to date `develop` branch, in `moduleconfs.data.xml`, we defined the following,

```
<obj class="TPCRawDataProcessor" id="def-wib-processor">
 <attr name="queue_sizes" type="u32" val="10000"/>
 <attr name="thread_names_prefix" type="string" val="postproc-"/>
 <attr name="latency_monitoring" type="bool" val="1"/>
 <attr name="channel_map" type="string" val="PD2HDTPCChannelMap"/>
 <attr name="metric_collect_opmon_rate" type="u32" val="1"/>
 <rel name="processing_steps">
  <ref class="AVXFrugalPedestalSubtractProcessor" id="tpg-pedsub-proc"/>
  <ref class="AVXThresholdProcessor" id="tpg-threshold-proc"/>
 </rel>
 <rel name="sot_minima" class="SamplesOverThresholdMinima" id="def-sot-minima"/>
</obj>
```

However, if we go to the resolved `integtest-session-resolved.data.xml`, we find the resolved configuration to be,

```
<obj class="TPCRawDataProcessor" id="def-wib-processor">
 <attr name="queue_sizes" type="u32" val="10000"/>
 <attr name="thread_names_prefix" type="string" val="postproc-"/>
 <attr name="latency_monitoring" type="bool" val="1"/>
 <attr name="channel_map" type="string" val="PD2HDTPCChannelMap"/>
 <attr name="metric_collect_opmon_rate" type="u32" val="1"/>
 <rel name="processing_steps">
  <ref class="AVXThresholdProcessor" id="tpg-threshold-proc"/>
 </rel>
 <rel name="sot_minima" class="SamplesOverThresholdMinima" id="def-sot-minima"/>
</obj>
```

`<ref class="AVXFrugalPedestalSubtractProcessor" id="tpg-pedsub-proc"/>` is missing from the resolved processing steps, giving us only one processor instead.

# Proposed solution

Appearing to be a configuration resolution issue, I have lifted the definitions of the processors from `ru-segment.data.xml` directly into `moduleconfs.data.xml`. In addition, I have also removed a redundant definition of the `AVXFrugalPedestalSubtractProcessor` from `integrationtest-objects.data.xml`.

# Tests

- [x] tpstream_writing_test
- [x] example_system_test
- [x] Full integration tests

### DIFF
--- a/config/daqsystemtest/integrationtest-objects.data.xml
+++ b/config/daqsystemtest/integrationtest-objects.data.xml
@@ -81,11 +81,6 @@
  <comment creation-time="20240909T081245" created-by="glehmann" created-on="np04-srv-024.cern.ch" author="glehmann" text=" "/>
 </comments>
 
-
-<obj class="AVXFrugalPedestalSubtractProcessor" id="tpg-pedsub-proc">
- <attr name="accum_limit" type="u8" val="10"/>
-</obj>
-
 <obj class="DFHWConf" id="dfhw-01">
  <rel name="uses">
   <ref class="StorageDevice" id="storage-01"/>
--- a/config/daqsystemtest/moduleconfs.data.xml
+++ b/config/daqsystemtest/moduleconfs.data.xml
@@ -89,12 +89,35 @@
  <comment creation-time="20250805T092423" created-by="dergonul" created-on="np04-srv-028.cern.ch" author="dergonul" text="Emu CRT configs"/>
 </comments>
 
-
 <obj class="AVXThresholdProcessor" id="tpg-threshold-proc">
  <attr name="metric_collect_time_sample_period" type="u64" val="256"/>
  <attr name="plane0" type="u16" val="100"/>
  <attr name="plane1" type="u16" val="100"/>
  <attr name="plane2" type="u16" val="100"/>
+</obj>
+
+<obj class="AVXAbsRunSumProcessor" id="tpg-absrs-proc">
+ <attr name="metric_collect_time_sample_period" type="u64" val="256"/>
+ <attr name="memory_factor_plane0" type="u8" val="8"/>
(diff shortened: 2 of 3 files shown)


bieryAtFnal (approved):
This change looks reasonable to me.  I verified that it fixes the missing processor entry in the TPCRawDataProcessor declaration in the generated configuration.  I ran all regression tests in the `daqsystemtest` and `dfmodules` repos. The only unexpected failure was in the 2x3 part of the `example_system_test`, in which the number of TriggerRecords was slightly low.  I was unable to reproduce that failure, though.

xinyue-uoft:
> This change looks reasonable to me. I verified that it fixes the missing processor entry in the TPCRawDataProcessor declaration in the generated configuration. I ran all regression tests in the `daqsystemtest` and `dfmodules` repos. The only unexpected failure was in the 2x3 part of the `example_system_test`, in which the number of TriggerRecords was slightly low. I was unable to reproduce that failure, though.

I guess a question to ask is whether the expected `TriggerRecords` is designed to fit the correct 2 processor case, or was it actually designed against the faulty 1 processor result?

bieryAtFnal:
The expected number of Trigger Records was set for the faulty 1-processor case.  The fact that that a similar number of TRs is created for the 2-processor case makes me wonder what the second (pedestal) processor is doing.  It is probably calculating pedestals and subtracting them from the ADC values, as the name suggests.  However, since the fake-data-replay file that we use most (all?) of the time for TPG in regression tests is `wibeth_output_all_zeros.bin`, the pedestal processor doesn't have much effect.

PR #249, opened 2025-10-16
Remove eth_device_name from base configurations

eflumerf:
Due to the changes in DUNE-DAQ/utilities#57, warning messages are now printed when a Service targets a non-existent `eth_device_name`. To that end, this PR removes the `eth_device_name` defined for Services in the common configuration.


roland-sipos (approved):

jcfreeman2:
Roland on Slack, 11:16 AM local time, re: merges: "Go for it John whenever it suits you."

PR #254, opened 2025-10-21
New integration test for tpg internal state monitor

xinyue-uoft:
Part of https://github.com/DUNE-DAQ/daq-deliverables/issues/178

Addes new unit test that is modified on top of current tpstream_writing_test. Enables tpg internal state monitoring.


aeoranday (approved):
LGTM. Integration tests passed and the new one added contained the additional metric values in the `info.*.json` file.

PR #255, opened 2025-10-21
Fixes for the TDEEth_TPG part of the readout_type_scan_test

bieryAtFnal:
## Description

We noticed that the recent addition of a TDEEth_TPG section to the `readout_type_scan_test` started to fail when it was run with recent changes in `fdreadoutlibs` (creation of `TPCEthFrameProcessor`).

Alejandro found that this was because of several reasons:

1. Kurt had used an invalid GeoID (crate/slot/stream == 5/3/1) for one of the data sources in the integtest.  Alejandro fixed this problem by using 5/4/0 instead of 5/3/1.
2. The sample TDE data had several noisy time periods in its data sample, and this caused the rate of TPs to be higher than expected.  This was fixed by tuning the TP threshold and TAMaker prescale.  

## Testing Suggestions

These changes can be tested by checking out this branch in a local software area and verifying that the `readout_type_scan_test` now runs cleanly.

### DIFF
--- a/integtest/readout_type_scan_test.py
+++ b/integtest/readout_type_scan_test.py
@@ -206,9 +206,23 @@ tde_tpg_conf.config_substitutions.append(
 )
 tde_tpg_conf.config_substitutions.append(
     data_classes.attribute_substitution(
+        obj_class="AVXThresholdProcessor",
+        obj_id="tpg-threshold-proc",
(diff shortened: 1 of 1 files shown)


bieryAtFnal:
Here is the plot from Alex of the waveform data in the TDE asset file...

tdeeth_asset_waveforms

PR #285, opened 2026-01-14
Proposed name change for the script that validates the example configs in daqsystemtest

bieryAtFnal:
# Description

Renamed the `check_daqsystemtest_applications.sh` as `check_example_configs.sh` to try to make its purpose more clear...


eflumerf (approved):
I looked at the script again, and really it's pretty generic except for the last 7 lines. Would there be any other places where the check_config function would be useful? (e.g. ehn1-daqconfigs)

bieryAtFnal:
> I looked at the script again, and really it's pretty generic except for the last 7 lines. Would there be any other places where the check_config function would be useful? (e.g. ehn1-daqconfigs)

I suspect yes.
In a similar vein, I was thinking that it would be easy, and possibly useful, to add command-line options to pick which (one or more) top-level configuration *.data.xml file(s) to check.
I'll file an Issue for that.

PR #290, opened 2026-01-28
Script for running integtests in whichever repos the user specifies

bieryAtFnal:
# Description

In recent discussions about the running of integtests (regression tests and integration tests), people (including me) mentioned several requests for enhancements. Those included the following:

- providing "_bundle_" script functionality in all repos that have integtests without adding the maintenance burden of having to maintain N different scripts
- the ability to run test from repos that have not been cloned into a local software area

This changes covered by this PR attempt to meet that need.  A global bundle script named `dunedaq_integtest_bundle.sh` has been created and a couple of utility scripts for finding repos with integtests and listing the integtests in a given repo are included.

A markdown document has been created in the `docs` subdirectory of this repo with an introduction to these new scripts.  Please refer to that document for instructions on how to use the scripts.

Testing of these changes would include running various combinations of integtests using the `dunedaq_integtest_bundle.sh` script and specifying different command-line options to customize the behavior of the script.

To get access to these scripts before they are merged to the `develop` branch of `daqsystemtest`, users should include the `kbiery/global_bundle_script` branch of `daqsystemtest` in their local software area(s) and rebuild the software area to get the scripts copied into the appropriate `bin` areas.

Please note that some of the command-line options that were available in the older `daqsystemtest_integtest_bundle.sh` script have not been carried forward to the new global bundle script.  For example, ones that specify the Nth thru Mth tests to be run.  That functionality became less useful in the global bundle script since the list of tests to be run is much more dynamic with that script.

## Type of change

- [x] Documentation (non-breaking change that adds or improves the documentation)

## Testing checklist


scripts/list_repos_with_integtests.sh
```diff
@@ -0,0 +1,49 @@
...
+    echo "Looking for _all_ repositories with integtests in them..." >&2
+    echo "" >&2
+
+    # determine the base release directory
+    release_dir=`dbt-info release | grep 'Release dir:' | cut -d' ' -f3`
+    base_release_name=`dbt-info release | grep 'Base release name:' | cut -d' ' -f4`
+    base_release_dir="${release_dir}/../${base_release_name}"
+
+    # look up the paths of all of the repositories in the core and detector-specific categories
+    det_rel_repo_paths=(`ls -1d ${release_dir}/spack-installation/opt/spack/*almalinux9*/gcc-*/*/*/integtest/*_test.py`)
```
eflumerf: AL10 will soon be a thing...It might be as simple as `*almalinux*` (though it may get duplicates that way)

eflumerf (approved):
Works as advertised.

scripts/list_repos_with_integtests.sh
bieryAtFnal: Good point.  The current path has `linux-almalinux9-x86_64` in that element, so I switched to just `linux`.  Hopefully that will be moderately robust going forward.  Thanks!

jcfreeman2:
I've played around with this and I think it looks great. The only thing I'd suggest is to provide a link to the script-specific page from the main page (`README.md`).

bieryAtFnal:
@jcfreeman2 , as you suggested, I have updated the top-level README.md file and the README.md file in the `integtest` subdir.  Please let me know if the new text looks reasonable.

jcfreeman2:
@bieryAtFnal I think it's great that not only did you provide the link but you overhauled the antiquated documentation. Approved, with one quibble, which is that the link to the OKS XML file should be changed to:
```
[this file](https://github.com/DUNE-DAQ/daqsystemtest/blob/develop/config/daqsystemtest/example-configs.data.xml).
```
...since otherwise the link will be broken in https://dune-daq-sw.readthedocs.io/en/latest/packages/daqsystemtest

jcfreeman2 (approved):
Approved; see my earlier comments in this PR for details.

bieryAtFnal:
Thanks, @jcfreeman2  - great catch on that bad link.

PR #292, opened 2026-02-05
Added checking of computer resources in integtests

bieryAtFnal:
# Description

Please note that, as of 13-Feb-2026, these changes are dependent on the ones in DUNE-DAQ/integrationtest#143.  The latest additions support separate checking of _recommended_ computer resources compared to _required_ computer resources.

To help avoid confusion when the `daqsystemtest` integtests are run on computers that may or may not have sufficient resources (CPUs, memory, disk space), I have added checks on such resources to all of the integtests in this repo.  When a computer does not have sufficient resources, the integtest is skipped.  (This is accomplished by having the `pytest` "tests" skipped using the `pytest.skip()` functionality.)

To test these changes, I have run the modified integtests (typically with `dbt-build --integtest`) on several computers, some of which have lots of resources, and some of which don't.  I considered the tests successful when either they ran cleanly or they were skipped because of insufficient resources.

It would be great for reviewers to also run these modified tests on different computers and verify similar behavior (consistent success or skipping).

## Type of change

## Testing checklist

## Further checks


eflumerf:
Looking through the changes, I see a lot of comments like "2x what we need".

In my past testing, I have had every test pass except for `3ru_1df` which already had skips in place (on a node with 32 CPUs and ~48G of memory). I'm wondering if we should have a "minimum required" and a "recommended" value, with a note that if you're not at "recommended" levels the test may fail, especially if the computer running the test is shared.

bieryAtFnal:
@eflumerf , ask you suggested/requested, there is now support for checking each of required computer resources and recommended computer resources.

In order to test these latest changes, we need to include the changes in DUNE-DAQ/integrationtest#143 in our software areas.

Please take a look and let me know what you think.  One question:  is a prefix of "*** Note:" for recommended-resource warnings sufficient, or should I use "\N{LARGE BLUE CIRCLE}"?  (I started with a large yellow circle, but decided that prefix made the problem seem more like an error than a warning.  Thanks

eflumerf (approved):

eflumerf:
I think it looks great as-is.

PR #308, opened 2026-03-24
Removed the unnecessary and confusing second tc-map entry from the de…

bieryAtFnal:
…f-tc-processor object in moduleconfs.data.xml.

# Description

This change is targeted to `fddaq-v5.7.0`...

It is very confusing and misleading to have two tc-map entries in a TCDataProcessor object that both specify the kTiming TC type.  Please see the snippets from `moduleconfs.data.xml` here:

```
<obj class="TCDataProcessor" id="def-tc-processor">
 <attr name="queue_sizes" type="u32" val="10000"/>
 <attr name="thread_names_prefix" type="string" val="mlt_"/>
 <attr name="latency_monitoring" type="bool" val="1"/>
 <attr name="merge_overlapping_tcs" type="bool" val="1"/>
 <attr name="ignore_overlapping_tcs" type="bool" val="0"/>
 <attr name="td_out_of_timeout" type="bool" val="1"/>
 <attr name="buffer_timeout" type="u32" val="1"/>
 <attr name="td_readout_limit" type="u32" val="1000000"/>
 <rel name="tc_readout_map">
  <ref class="TCReadoutMap" id="def-tc-map"/>
  <ref class="TCReadoutMap" id="def-hsi-tc-map"/>
 </rel>
</obj>

<obj class="TCReadoutMap" id="def-hsi-tc-map">
 <attr name="tc_type_name" type="string" val="kTiming"/>
 <attr name="time_before" type="u32" val="3000"/>
 <attr name="time_after" type="u32" val="1001"/>
</obj>

<obj class="TCReadoutMap" id="def-tc-map">
 <attr name="tc_type_name" type="string" val="kTiming"/>
 <attr name="time_before" type="u32" val="3000"/>
 <attr name="time_after" type="u32" val="1001"/>
</obj>
```

I propose to remove one of them.  In this PR, I have removed the `def-tc-map` entry and kept the `def-hsi-tc-map` entry, but I'll be glad to reverse that choice if the opposite choice makes more sense.

I have run the full set of `daqsystemtest` and `dfmodules` integtests with this change and seen no problems related to the change.

I found this duplication when I attempted to modify the readout window in an integtest (using the configuration-substitution functionality) and it wasn't clear which entry to change.  (And changing the first one in the list had no effect!)

## Type of change

## Testing checklist

### DIFF
--- a/config/daqsystemtest/moduleconfs.data.xml
+++ b/config/daqsystemtest/moduleconfs.data.xml
@@ -815,7 +815,6 @@
  <attr name="buffer_timeout" type="u32" val="1"/>
  <attr name="td_readout_limit" type="u32" val="1000000"/>
  <rel name="tc_readout_map">
-  <ref class="TCReadoutMap" id="def-tc-map"/>
   <ref class="TCReadoutMap" id="def-hsi-tc-map"/>
  </rel>
 </obj>
--- a/config/daqsystemtest/moduleconfs.data.xml
+++ b/config/daqsystemtest/moduleconfs.data.xml
@@ -841,12 +841,6 @@
  <attr name="time_after" type="u32" val="1001"/>
 </obj>
 
-<obj class="TCReadoutMap" id="def-tc-map">
- <attr name="tc_type_name" type="string" val="kTiming"/>
- <attr name="time_before" type="u32" val="3000"/>
- <attr name="time_after" type="u32" val="1001"/>
-</obj>
-
 <obj class="TCReadoutMap" id="prescale-tc-map-entry">
  <attr name="tc_type_name" type="string" val="kPrescale"/>
  <attr name="time_before" type="u32" val="2000"/>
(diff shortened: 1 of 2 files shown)


alessandrothea (approved):
Agreed. 
Do we want to keep the `def-tc-map` object or should it go as well?

aeoranday:
If `def-tc-map` is to be removed entirely, then it should also be removed from `integtest/trigger_bitwords_test.py`. It looks like it would be safe to do so (and it's unclear what it's accomplishing by being there).

bieryAtFnal:
At first, I was thinking that we could keep `def-tc-map` around in case it will be useful in the future.  But, I'm now thinking that it could easily be added back if/when there is a need for it.  So, I've removed it.  And, my thanks to Alex for pointing out the need to remove it from that one integtest.

PR #325, opened 2026-06-24
Modified integtests to make use of new utility functions for setting trigger config params

bieryAtFnal:
# Description

These changes make use of the changes made in the `integrationtest` repo to provide utility functions that set various trigger parameters for integration tests (DUNE-DAQ/integrationtest#161).

There are instructions for testing these changes in DUNE-DAQ/integrationtest#161, and these changes need to be tested and merged together the with changes in the `integrationtest` repo.

## Type of change

## Testing checklist

### DIFF
--- a/integtest/readout_type_scan_test.py
+++ b/integtest/readout_type_scan_test.py
@@ -180,7 +180,7 @@ conf_dict.op_env = "integtest"
 conf_dict.config_session_name = "readout"
 conf_dict.tpg_enabled = False
 conf_dict.frame_file = "asset://?label=ProtoWIB&subsystem=readout"  # ProtoWIB
-utility_functions.set_RTCM_trigger_params(conf_dict, trigger_rate=1)
+utility_functions.set_rtcm_trigger_params(conf_dict, trigger_rate=1)
 
 wib_tpg_conf = copy.deepcopy(conf_dict)
 wib_tpg_conf.tpg_enabled = True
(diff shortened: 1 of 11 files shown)


integtest/trigger_bitwords_test.py
```diff
@@ -200,30 +200,12 @@
...
-        obj_id="def-hsi-tc-map",
-        obj_class="TCReadoutMap",
-        updates={"time_before": 62500, "time_after": 62500},)
-)
+utility_functions.set_RTCM_trigger_params(coincidence_bitword_conf, trigger_rate=40,
+                                          readout_window_backshift_ticks=0, readout_window_before_ticks=62500,
+                                          readout_window_after_ticks=62500)
+utility_functions.set_fake_hsi_trigger_params(coincidence_bitword_conf, trigger_rate=30,
+                                              readout_window_before_ticks=62500,
+                                              readout_window_after_ticks=62500)
```
MRiganSUSX: The 62500-tick readout window for kRandom triggers in coincidence_bitword_conf is now set only on RandomTCMakerConf via set_RTCM_trigger_params. The TCReadoutMap entry def-random-readout — which exists in the template XML — is left at its defaults of 1000/1001 ticks. This is safe today because def-random-readout is not currently referenced in the MLT tc_readout_map relation. But if it ever is added there, TCReadoutMap supersedes, and the test's readout window would silently collapse from 62500 to 1000 ticks.

TLDR, this is one of _those_ places where things work atm, but are relying on assumptions and choices made in the default base config but can break. 
(not sure what to do about that, I'm just thinking out loud here)

integtest/readout_type_scan_test.py
MRiganSUSX: should these two be migrated to new functions as well ?

MRiganSUSX (commented):
Few small comments here

integtest/trigger_bitwords_test.py
bieryAtFnal: Yes.  There are other integtests that have this same feature, such as the `long_window_readout_test`.  As you say, if something in the underlying configuration or code changes, then the tests will need to change also.

integtest/readout_type_scan_test.py
bieryAtFnal: Yes, and I've now done this.

MRiganSUSX (approved):
LGTM.

PR #326, opened 2026-07-01
Modified the bundle script and its helper scripts to include integtests from python-only packages

bieryAtFnal:
# Description

Now that we have integration tests from Python packages installed in the `.venv` directory of software areas, we can proceed to include them in the global bundle script and its helper scripts:

- `dunedaq_integtest_bundle.sh`
- `list_available_integtests.sh`
- `list_repos_with_integtests.sh`

The changes in this Pull Request attempt to do that.

Two notes:

- I have attempted to continue to support the "local" keyword for finding integtests in software packages that are included in a local software area.  For Python packages, this corresponds to the presence of the package in the `pythoncode` subdirectory.  (Typically, **_all_** Python packages are installed in the `.venv` subdirectory of a local software area, so the presence of a package in that area can't be used to determine "local" vs. not-local.)
- I have **_not_** yet added logic to the bundle script to add the `-p no:cacheprovider` qualifier to the `pytest` command when running integtests from Python packages that are not installed in a write-able location on disk.  The reason for this is twofold: there are discussions about how to avoid needing to specify that qualifier, so we will let that discussion proceed.  And in the short term, I expect that it will be rather rare for someone to run integtests from Python packages without having a local software area.

Here are suggested steps for testing these changes:

```
DATE_PREFIX=`date '+%d%b'`
TIME_SUFFIX=`date '+%H%M'`

setup_dbt latest
dbt-create -n NFD_DEV_260701_A9 ${DATE_PREFIX}FDDevBundleScriptPythonPkgs_${TIME_SUFFIX}
cd ${DATE_PREFIX}FDDevBundleScriptPythonPkgs_${TIME_SUFFIX}/sourcecode

git clone https://github.com/DUNE-DAQ/daqsystemtest.git -b kbiery/python_package_integtests
cd ..

. ./env.sh
dbt-build -j 12
dbt-workarea-env

dunedaq_integtest_bundle.sh -r all --list

echo ""
echo -e "\U1F535 \U2705 Note that the drunc integtest is included in the list of *all* tests which would be run. \U2705 \U1F535"
echo ""
echo ""
sleep 3

dunedaq_integtest_bundle.sh -r local --list

echo ""
echo -e "\U1F535 \U2705 Note that the drunc integtest is *not* included in the list of *local* tests which would be run. \U2705 \U1F535"
echo ""
echo ""
sleep 3

cd pythoncode
git clone https://github.com/DUNE-DAQ/drunc.git -b develop
cd ..

dunedaq_integtest_bundle.sh -r local --list

echo ""
echo -e "\U1F535 \U2705 Note that the drunc integtest *is now included* in the list of *local* tests which would be run. \U2705 \U1F535"
echo ""
echo ""
sleep 3

dunedaq_integtest_bundle.sh -r drunc

echo ""
echo -e "\U1F535 \U2705 Note that the drunc process_manager_test ran successfully using the bundle script. \U2705 \U1F535"
echo ""
echo ""
```

## Type of change

## Testing checklist

## Further checks

### DIFF
--- a/scripts/dunedaq_integtest_bundle.sh
+++ b/scripts/dunedaq_integtest_bundle.sh
@@ -285,7 +285,10 @@ while [[ ${full_set_loop_count} -lt ${full_set_requested_interations} ]]; do
             echo -e "\U0001F535 \033[0;34mStarting test ${overall_test_index} of ${total_number_of_tests}...\033[0m \U0001F535" | CaptureOutput ${ITGRUNNER_LOG_FILE}
 
             echo -e "\u2B95 \033[0;1mRunning ${FULL_TEST_NAME}\033[0m \u2B05" | CaptureOutput ${ITGRUNNER_LOG_FILE}
-            if [[ -e "./${test_name}" ]]; then
+
+            if [[ "`ls ${DBT_AREA_ROOT}/.venv/lib/python*/site-packages/${test_repo}/integtest/${test_name} 2>/dev/null`" != "" ]]; then
+                ${PYTEST_COMMAND} ${DBT_AREA_ROOT}/.venv/lib/python*/site-packages/${test_repo}/integtest/${test_name} | CaptureOutputNoANSI ${ITGRUNNER_LOG_FILE}
+            elif [[ -e "./${test_name}" ]]; then
                 ${PYTEST_COMMAND} ./${test_name} | CaptureOutputNoANSI ${ITGRUNNER_LOG_FILE}
             elif [[ -e "${DBT_AREA_ROOT}/sourcecode/${test_repo}/integtest/${test_name}" ]]; then
                 if [[ -w "${DBT_AREA_ROOT}" ]]; then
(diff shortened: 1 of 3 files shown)


jcfreeman2:
I checked the given commands and they work as advertised as described in the above script. 

One issue I bumped into, however - the line:
"repo_name=\`echo ${path} | cut -d'/' -f 11\`"
from `list_repos_with_integtests.sh` has a hardwired assumption (`-f 11`) about where the name of the `drunc` repo is in the path. This is effectively an assumption about the number of subdirectories in `${DBT_AREA_ROOT}` which isn't guaranteed to hold.

bieryAtFnal:
I have modified the `list_repos_with_integtests.sh` script to more reliably determine whether a Python package is "local" or not. This addresses the problem that John noticed (thanks!)

jcfreeman2:
I've confirmed both just from looking at the `sed` commands and from rerunning `list_repos_with_integtests.sh` that the problem is now fixed; approved.

jcfreeman2 (approved):
See comment I left a moment ago.

PR #335, opened 2026-08-18
Modified the 3ru_3df_multirun_test to make use of the 'start-run' command...

bieryAtFnal:
…for one of its three runs. This will validate the use of that command going forward.

# Description

At today's SWI&T meeting, Michal mentioned that he noticed that the `start-run` command in run control was running `boot` and `conf` when those transitions had already been run in the current DAQ session.

Surprisingly, we don't seem to have an integration test that uses `start-run`, but we certainly should.

This PR makes use of that command in one of the three data-taking runs that are part of the `3ru_3df_multirun_test`.

The modified version of this integtest will (and **_should_**) fail until the bug is found and fixed in run control.

## Type of change

## Testing checklist

- [ ] Full set of integration tests pass (`dunedaq_integtest_bundle.sh`)
    - NB, this test will fail initially and that is a good thing. The failure demonstrates a problem in run control.

### DIFF
--- a/integtest/3ru_3df_multirun_test.py
+++ b/integtest/3ru_3df_multirun_test.py
@@ -176,7 +176,7 @@ dunerc_command_list += (
     + "disable-triggers wait 1 drain-dataflow wait 2 stop-trigger-sources wait 1 stop wait 2".split()
 )
 dunerc_command_list += (
-    "start --run-number 103 wait 1 enable-triggers wait ".split()
+    "start-run --run-number 103 wait ".split()
     + [str(run_duration)]
     + "disable-triggers wait 1 drain-dataflow wait 2 stop-trigger-sources wait 1 stop wait 2".split()
 )


MRiganSUSX (approved):
Thank @bieryAtFnal.

Can confirm that:
- running `3ru_3df_multirun_test` the errors are caught

Image

Image

Image

(although not directly mentioned in this `Failures` end block (?))

- running with drunc @ https://github.com/DUNE-DAQ/drunc/pull/1015 which is the fix targetting this, there are no more errors

Image

bieryAtFnal:
@MRiganSUSX , thanks for your review.  I'm going to go ahead and merge these changes to the `develop` branch, and I've created an Issue in the `integrationtest` repo to request that we make the text in the 'failure end block' more useful (DUNE-DAQ/integrationtest#170).

